PR #5, opened 2024-06-02
Initial Commit with CLI parsing

sridamul:
In this PR,
- I set up the project
- Implemented the CLI module functions (not all)
- Unit test coverage for CLI module classes
- Integrated spotbugs
- Used maven assembly plugin to build jar with dependencies

cc: @berviantoleo, @iamrajiv, @jonesbusy, @gounthar


gh-integration/pom.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.jenkins.plugin-modernizer</groupId>
+        <artifactId>plugin-modernizer-parent-pom</artifactId>
+        <version>${revision}${changelist}</version>
+        <relativePath>../pom.xml</relativePath>
```
jonesbusy, line 10: If i'm not mistaken the relativePath is optional if found of direct parent directory

core/pom.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.jenkins.plugin-modernizer</groupId>
+        <artifactId>plugin-modernizer-parent-pom</artifactId>
+        <version>${revision}${changelist}</version>
```
jonesbusy, line 9: If we are using CD also for the tool (Which I like very much) I think we are missing some files like https://www.jenkins.io/doc/developer/publishing/releasing-cd/#update-maven-pom-and-config

Also incrementals must be enabled (I would suggest to check some plugin with CD enabled)

sridamul: Will take a look into it

core/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>core</artifactId>
```
jonesbusy, line 13: I think using a suffix is more clear for the artifact ID of child module

`plugin-modernizer-parent-core`
`plugin-modernizer-parent-cli`

sridamul: is `plugin-modernizer-core, plugin-modernizer-cli, plugin-modernizer-recipes, plugin-modernizer-integration` ok? (or) `parent` should be added to each module?

jonesbusy: Looks good.  `parent` (or `pom`) is only for the top level pom module

core/pom.xml
sridamul: Hey! @jonesbusy I have enabled incrementals and added Dependabot. Will configure others later.

jonesbusy (approved):

PR #15, opened 2024-06-11
Extract metadata from pom.xml file

sridamul:
This PR extracts essential metadata from pom.xml

Extraction of Metadata mentioned in [this list](https://docs.google.com/document/d/1D12RVhSrg0WTVCfCTcIE9zAhlL_YeydF59e7NLFQWQY/edit#heading=h.qdoxsfffxxxe) are implemented.

### Submitter checklist


jonesbusy:
Why we are not using OpenRewrite to extract metadata ? And checks existence of markers

```java
@Value
@EqualsAndHashCode(callSuper = false)
public class IsJenkinsPlugin extends Recipe implements IRecipe {

    ....

    @Override
    public TreeVisitor<?, ExecutionContext> getVisitor() {
        return new XmlVisitor<ExecutionContext>() {
            @Override
            public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
                // TODO
            }
        };
    }

    ...

}
```

My opinion is that OpenRewrite is good at parsing pom.xml and the full hierarchy of dependency

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/Developer.java
```diff
@@ -0,0 +1,14 @@
+package io.jenkins.tools.pluginmodernizer.core.extractor;
+
+public record Developer(String id, String name, String email, String role) {
```
jonesbusy, line 3: Developer section are not used anymore on Jenkins pom.xml.

In fact that even could be a receipt to cleanup such pom

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/PomParser.java
```diff
@@ -0,0 +1,182 @@
...
+
+public class PomParser {
+
+    private static final Logger logger = Logger.getLogger(PomParser.class.getName());
+
+    private final Document doc;
+    private final XPath xpath;
+    private final Map<String, String> properties;
+
+    public PomParser(String filePath) {
```
jonesbusy, line 27: OpenRewrite will do the parsing for you. I don't think that's the goal to implement a custom parser.

PR #16, opened 2024-06-16
Extract metadata from pom.xml

sridamul:
Initial Iteration to extract essential metadata from pom.xml
### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/PluginMetadata.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/PluginMetadata.java
@@ -3,15 +3,104 @@ package io.jenkins.tools.pluginmodernizer.core.extractor;
 import java.util.List;
 import java.util.Map;
 
-import org.openrewrite.maven.tree.*;
+import org.openrewrite.maven.tree.Dependency;
+import org.openrewrite.maven.tree.Parent;
 
 public class PluginMetadata {
     private String pluginName;
-    private String pluginVersion;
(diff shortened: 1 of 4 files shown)


jonesbusy:
To fix spotbug and spotless

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/PluginMetadata.java
```diff
@@ -0,0 +1,17 @@
+package io.jenkins.tools.pluginmodernizer.core.extractor;
+
+import java.util.List;
+import java.util.Map;
+
+import org.openrewrite.maven.tree.*;
+
+public class PluginMetadata {
```
jonesbusy, line 8: I would keep it very very simple for now.

Probably a first iteration would be

pluginName
parentVersion
jenkinsVersion

What do you think ?

The rest could be added later if we need to

sridamul:
```    
    private String pluginName;
    private boolean isLicensed;
    private boolean hasDevelopersTag;
    private boolean hasJavaLevel;
    private boolean usesHttps;
    private List<Dependency> dependencies;
    private String jenkinsVersion;
    private Parent pluginParent;
    private Map<String, String> properties;
```

Already extracted the things in this list. Will commit after resolving spotbugs and checkstyle issues.

jonesbusy:
Thanks, lets merge it as is and improve in the future

jonesbusy (approved):

PR #20, opened 2024-06-21
Invoke OpenRewrite Maven Plugin to execute recipes

sridamul:
Updates...
1) Now we can pass the recipe names as described in the [json file](https://github.com/jenkinsci/plugin-modernizer-tool/compare/main...sridamul:core-iteration-1?expand=1#diff-cbe860447dc0719f7f4a6ddcc831181e8eb4fb33fed0a3078a2939735ced7363), which resolves the recipe artifact coordinates.
2) PomParser recipe generates a metadata.json file in the target directory (I did this because the fetched plugin metadata is in new JVM which cannot be accessed. If can please let me know)
3) We can now run the recipes in multiple plugins (right now I tested by hardcoding two plugins in the test-plugins directory as shown in the video, in future PRs we can populate the directory with plugins)

https://github.com/jenkinsci/plugin-modernizer-tool/assets/107404972/269899cf-9eb4-4f1d-84f0-9b15e25387c6

Note: The error show in the video is due to SSLHandshakeException (I haven't added the certificate in my local env) which doesn't affect the execution of recipes

Manual testing done.

### Submitter checklist


plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
```diff
@@ -1,23 +1,142 @@
...
+                LOG.info("Build succeeded.");
+            }
+        } catch (MavenInvocationException e) {
+            LOG.error("Maven invocation failed: ", e);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static String getMavenHomePath() {
```
jonesbusy, line 98: This must be extracted to Settings.java https://github.com/jenkinsci/plugin-modernizer-tool/pull/18/files#diff-c2c9a97539ad9212e10cf7f95dcc8b8374c3db445be1f14d9edd8595d14ae004R6 to be used in order places

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
```diff
@@ -1,23 +1,142 @@
...
+        String mavenHome = System.getenv("MAVEN_HOME");
+        if (mavenHome == null) {
+            mavenHome = System.getenv("M2_HOME");
+        }
+        return mavenHome;
+    }
+
+    List<String> createGoalsList() throws IOException {
+        List<String> goals = new ArrayList<>();
+        goals.add("org.openrewrite.maven:rewrite-maven-plugin:run");
```
jonesbusy, line 108: I think using specific version of the plugin is important to avoid breaking changes. Could be just a constant inside settings

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
```diff
@@ -1,23 +1,142 @@
...
+
+        try (InputStream inputStream = PluginModernizer.class.getResourceAsStream("/recipe_data.json")) {
```
jonesbusy, line 110: I would extract this logic a a separate class that PluginModernizer.java is using. 

It the the PluginModernizer.java will just call using and orchestrate the modernizing of plugin

jonesbusy (changes requested):
Thanks! Looks good overall, some small cosmetic things and extraction to separate class

plugin-modernizer-core/src/main/resources/recipe_data.json
```diff
@@ -0,0 +1,12 @@
+{
```
jonesbusy, line 1: Is it something that we need to maintain manually or is is expected to be computed automatically ?

sridamul: right now, to check the working of the function I hard coded it, which will be automated later

jonesbusy: Sure perfect thanks, for the explanation

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/config/Config.java
```diff
@@ -10,12 +10,16 @@ public class Config {
     private final List<String> plugins;
     private final List<String> recipes;
     private final Path cachePath;
+    private final String mavenHome;
+    private final String mavenPluginVersion;
```
jonesbusy, line 14: mavenRewritePluginVersion is more explicit and clear that is the version of the rewrite maven plugin

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
```diff
@@ -31,8 +28,11 @@ public class PluginModernizer {
 
     private static final ObjectMapper objectMapper = new ObjectMapper();
 
+    private final GoalsListCreator goalsListCreator;
+
     public PluginModernizer(Config config) {
         this.config = config;
+        this.goalsListCreator = new GoalsListCreator(config);
```
jonesbusy: Why not just something like creating a class responsible of maven invocation ?

I would expect to see a `start()` method of the `PluginModernizer` very simple where all the logic is encapsulated

Just an example of pseudo like (without any optimization or threading nor error handling)

```java
public void start() {
     for (String plugin : config.getPlugins()) {
            mavenInvoker.fetchMetadata(plugin)
            mavenInvoker.invokeGoal('clean')
            mavenInvoker.invokeRewrite()
     }
}
```

What do you think about trying to extract responsibility on different classes ?

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenGoalInvoker.java
```diff
@@ -0,0 +1,109 @@
...
+            for (String recipe : recipes) {
+                JsonNode recipeNode = recipesNode.get(recipe);
+                if (recipeNode != null) {
+                    String fqcn = recipeNode.get("fqcn").asText();
+                    activeRecipes.add(fqcn);
+
+                    String artifactCoordinates = recipeNode.get("artifactCoordinates").asText();
+                    recipeArtifactCoordinates.add(artifactCoordinates);
+                } else {
+                    System.err.println("Recipe '" + recipe + "' not found.");
```
jonesbusy, line 60: Use logger and not println statement

plugin-modernizer-core/src/main/resources/recipe_data.json
```diff
@@ -0,0 +1,12 @@
...
+  "recipes": {
+    "PomParser": {
+      "fqcn": "io.jenkins.tools.pluginmodernizer.core.extractor.PomParser",
+      "artifactCoordinates": "io.jenkins.plugin-modernizer:plugin-modernizer-core:999999-SNAPSHOT"
+    },
+    "AddPluginsBom": {
+      "fqcn": "org.openrewrite.jenkins.AddPluginsBom",
+      "artifactCoordinates": "org.openrewrite.recipe:rewrite-jenkins:0.8.1"
```
jonesbusy, line 9: Something that I was wondering if it make sense to create an uber jar that contains all their recipes and their dependencies and always consume them from plugin-modernizer-core. To be discussed later

plugin-modernizer-core/src/main/resources/recipe_data.json
```diff
@@ -0,0 +1,12 @@
+{
+  "recipes": {
+    "PomParser": {
+      "fqcn": "io.jenkins.tools.pluginmodernizer.core.extractor.PomParser",
+      "artifactCoordinates": "io.jenkins.plugin-modernizer:plugin-modernizer-core:999999-SNAPSHOT"
+    },
+    "AddPluginsBom": {
```
jonesbusy, line 7: Improvement also for later is to always use FQDN for recipes, not only the simple class name

jonesbusy (changes requested):

plugin-modernizer-core/src/main/resources/recipe_data.json
sridamul: seems like a good idea

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
```diff
@@ -17,38 +13,20 @@ public class PluginModernizer {
...
-                    mavenGoalInvoker.invokeMavenGoals(pluginDir, goals);
-                } catch (IOException e) {
-                    LOG.error(e.getMessage(), e);
-                }
-            }
-        } else {
-            LOG.error("No subdirectories found in {}", pluginsDir.getAbsolutePath());
+        for (String plugin : config.getPlugins()) {
+            String pluginPath = projectRoot + "/test-plugins/" + plugin;
+            mavenInvoker.invokeGoal(pluginPath, "clean");
```
jonesbusy, line 27: This was just an example, but let's go that way ;) Not sure if a clean phase is needed, but let's got for it

jonesbusy (approved):

sridamul:
@jonesbusy Can you create a  general GH discussion to clarify quick questions?

jonesbusy:
@sridamul Sure I will. Thanks

PR #30, opened 2024-06-22
Remove unused config and override maven home and cache on CLI

jonesbusy:
I don't think we need to change the mavenRewritePluginVersion at runtime. Can be just a constant on the `Settings`

But passing the `--maven-home` sounds more useful on environment that doesn't declare (M2_HOME or MAVEN_HOME)

Also use `Path`instead of using `String` for paths

### Submitter checklist


jonesbusy:
Draft need to fix tests

jonesbusy:
This is ready for review. The last commit also ensure to have a predictable version of maven when running integration tests (in the future) and not rely on environment but surefire env vars

sridamul (approved):
LGTM! Thanks!

PR #46, opened 2024-06-24
Add list recipe option

### LINKED ISSUE #24: Add new command line option - list recipes

Command line option to list the available recipes with one line description of what it does.

### Upstream changes

_No response_

Yes!

sridamul:
Closes #24 
Added new option to list the available recipes
### Submitter checklist

### DIFF
--- a/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
+++ b/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
@@ -1,7 +1,9 @@
 package io.jenkins.tools.pluginmodernizer.cli;
 
+import java.io.InputStream;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Map;
 
 import io.jenkins.tools.pluginmodernizer.core.config.Config;
 import io.jenkins.tools.pluginmodernizer.core.config.Settings;
@@ -9,6 +11,7 @@ import io.jenkins.tools.pluginmodernizer.core.impl.PluginModernizer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.bridge.SLF4JBridgeHandler;
+import org.yaml.snakeyaml.Yaml;
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Option;
@@ -24,7 +27,7 @@ public class Main implements Runnable {
     private static Logger LOG = LoggerFactory.getLogger(Main.class);
 
     public static void main(final String[] args) {
-        new CommandLine(new Main()).execute(args);
+        new CommandLine(new Main()).setOptionsCaseInsensitive(true).execute(args);
     }
 
     @Option(names = {"-p", "--plugins"}, required = true, description = "List of Plugins to Modernize.", split = ",", parameterConsumer = CommaSeparatedParameterConsumer.class)
@@ -44,6 +47,9 @@ public class Main implements Runnable {
 
     @Option(names = {"-m", "--maven-home"}, description = "Path to the Maven Home directory.")
     public Path mavenHome = Settings.DEFAULT_MAVEN_HOME;
+
+    @Option(names = {"-l", "--list-recipes"}, help = true, description = "List available recipes.")
+    public boolean listRecipes;
 
     public Config setup() {
         Config.DEBUG = debug;
@@ -66,8 +72,25 @@ public class Main implements Runnable {
         }
     }
 
+    public void listAvailableRecipes() {
+        Yaml yaml = new Yaml();
+        InputStream inputStream = getClass().getClassLoader().getResourceAsStream("recipe_data.yaml");
+        if (inputStream == null) {
+            LOG.error("Failed to load the available recipes.");
+            return;
+        }
+        Map<String, Object> data = yaml.load(inputStream);
+        LOG.info("Available recipes:");
+        Map<String, Map<String, String>> recipes = (Map<String, Map<String, String>>) data.get("recipes");
+        recipes.keySet().forEach(recipe -> LOG.info(recipe));
+    }
+
     @Override
     public void run() {
+        if (listRecipes) {
+            listAvailableRecipes();
+            return;
+        }
         LOG.info("Starting Plugin Modernizer");
         PluginModernizer modernizer = new PluginModernizer(setup());
         modernizer.start();


plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
```diff
@@ -66,8 +72,25 @@ public String getVersion() {
         }
     }
 
+    public void listAvailableRecipes() {
+        Yaml yaml = new Yaml();
+        InputStream inputStream = getClass().getClassLoader().getResourceAsStream("recipe_data.yaml");
```
jonesbusy, line 77: Can we move this constant to Settings.java ? I think were are few other place where it's used

plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
```diff
@@ -66,8 +72,25 @@ public String getVersion() {
...
+        if (inputStream == null) {
+            LOG.error("Failed to load the available recipes.");
+            return;
+        }
+        Map<String, Object> data = yaml.load(inputStream);
+        LOG.info("Available recipes:");
+        Map<String, Map<String, String>> recipes = (Map<String, Map<String, String>>) data.get("recipes");
```
jonesbusy, line 84: Would be nice to have some kind of model for this. Using such raw type make the code difficult to read

For example 

```Java
List<Recipe> recipes = (List<Recipe>) data.get("recipes");
```

Is much more easy to understand.

So if the format change we can change the model directly and not dig into the code to find map of map of string

Can you try to introduce a new class that will represent the Java structure of a recipe in `recipe_data.yaml` ?

I think also the type of object is not ideal inside `recipe_data.yaml` A list of recipe would be better that an map of recipes. What do you think ?

jonesbusy (changes requested):

plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
sridamul: Understood. Let me create a model.

sridamul:
Hey @jonesbusy, I think the new yaml file format is easy to read and parse. wdyt?

sridamul:
But I am not sure to have the new two classes in the model directory.

sridamul:
Added description.Any thoughts?
```
Available recipes:               
FetchMetadata - Fetches plugin metadata and stores in /target/pluginMetadata.json 
AddPluginsBom - Adds Jenkins plugins BOM at the latest release if the project depends on any managed versions or an outdated BOM is present. 
AddCodeOwner - Adds the {artifactId}-plugin-developers team to all files in .github/CODEOWNERS if absent. 
```

jonesbusy:
> Hey @jonesbusy, I think the new yaml file format is easy to read and parse. wdyt?
> 
> Old:
> 
> ```
> recipes:
>   FetchMetadata:
>     fqcn: "io.jenkins.tools.pluginmodernizer.core.extractor.MetadataCollector"
>     artifactCoordinates: "io.jenkins.plugin-modernizer:plugin-modernizer-core:999999-SNAPSHOT"
>   AddPluginsBom:
>     fqcn: "org.openrewrite.jenkins.AddPluginsBom"
>     artifactCoordinates: "org.openrewrite.recipe:rewrite-jenkins:0.8.1"
>   AddCodeOwner:
>     fqcn: "org.openrewrite.jenkins.github.AddTeamToCodeowners"
>     artifactCoordinates: "org.openrewrite.recipe:rewrite-jenkins:0.8.1"
> ```
> 
> New:
> 
> ```
> recipes:
>   - name: "FetchMetadata"
>     fqcn: "io.jenkins.tools.pluginmodernizer.core.extractor.MetadataCollector"
>     artifactCoordinates: "io.jenkins.plugin-modernizer:plugin-modernizer-core:999999-SNAPSHOT"
>   - name: "AddPluginsBom"
>     fqcn: "org.openrewrite.jenkins.AddPluginsBom"
>     artifactCoordinates: "org.openrewrite.recipe:rewrite-jenkins:0.8.1"
>   - name: "AddCodeOwner"
>     fqcn: "org.openrewrite.jenkins.github.AddTeamToCodeowners"
>     artifactCoordinates: "org.openrewrite.recipe:rewrite-jenkins:0.8.1"
> ```

Any reason to keep the root element "recipes" ? if this document doesn't contains other data it could be a list directly ? (Example is ansible and tasks: https://github.com/ansible/ansible-examples/blob/master/tomcat-memcached-failover/roles/lb-nginx/tasks/main.yml)

I will continue my review tomorrow, thanks for the good work anyway!

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/config/Settings.java
```diff
@@ -11,6 +12,8 @@ public class Settings {
 
     public static final String MAVEN_REWRITE_PLUGIN_VERSION = "5.34.1";
 
+    public static final InputStream RECIPE_DATA_YAML = Settings.class.getClassLoader().getResourceAsStream("recipe_data.yaml");
```
jonesbusy, line 15: I think it's dangerous to store the InputStream here. I think it's safer to only store the string.

It's public, so you don't know if caller will do with it (close it ?)

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/model/RecipeData.java
```diff
@@ -0,0 +1,18 @@
+package io.jenkins.tools.pluginmodernizer.core.model;
+
+import java.util.List;
+
+public class RecipeData {
```
jonesbusy, line 5: See https://github.com/jenkinsci/plugin-modernizer-tool/pull/46#discussion_r1651195402

Would avoid probably using `RecipeData`

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/model/RecipeDescriptor.java
```diff
@@ -0,0 +1,40 @@
+package io.jenkins.tools.pluginmodernizer.core.model;
+
+public class RecipeDescriptor {
```
jonesbusy, line 3: What about creating a java Record ?

jonesbusy (changes requested):

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenInvoker.java
```diff
@@ -79,7 +80,7 @@ private List<String> createGoalsList() throws IOException {
         return goals;
     }
 
-    private List<String> getActiveRecipes(List<String> recipes, JsonNode recipesNode) {
+    private List<String> getActiveRecipes(List<String> recipes, ArrayNode recipesNode) {
```
jonesbusy, line 83: Now that we have `RecipeDescriptor` why we don't use where it should instead of raw Jackson types ?

Also from what I understand from this method is that it return the list of coordinate of the active recipes ?

Shouldn't be called `getRecipesCoordinates` or something similar ?

Also using record is much more clear for a reader to understand

```java
private List<String> getRecipesCoordinates(List<RecipeDescriptor> availableRecipes,
                                           List<String> activeRecipes) {
```

For me I can be improved in a next PR

jonesbusy (approved):

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenInvoker.java
sridamul: Hey @jonesbusy,

> Also from what I understand from this method is that it return the list of coordinate of the active recipes ?

[getRecipesCoorinadates()](https://github.com/jenkinsci/plugin-modernizer-tool/blob/main/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenInvoker.java#L117) returns the list of coordinate of active recipes, while [getActiveRecipes()](https://github.com/jenkinsci/plugin-modernizer-tool/blob/main/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenInvoker.java#L104) returns the list of fqcn of recipes.

I have raised PR #71 to use record instead of raw Jackson types

Issue #48, opened 2024-06-25
Experiment with `jgit`, and keep it up to date


gounthar:
### What feature do you want to see added?

During our last meeting, we discussed the use of [`jgit`](https://github.com/eclipse-jgit/jgit). I'd like to propose a few ideas for consideration:

1. Version Alignment:
   What if we use the same version of `jgit` as the one used in either:
   - The [Jenkins Git plugin](https://plugins.jenkins.io/git)
   - The [Git Client Plugin](https://plugins.jenkins.io/git-client)

2. Version Discovery:
   We could utilize [updatecli](https://www.updatecli.io/docs/plugins/resource/maven/) to help us identify the correct version from the Git Client Plugin's pom file.

3. Dependency Management:
   We might consider adding the Git Client Plugin's pom as a Bill of Materials (BOM) for better dependency management.

4. Expert Input:
   @jonesbusy may have valuable insights or alternative approaches to suggest.

5. Initial Approach:
   For now, we could start with a pinned version of `jgit`. Later, we can explore ways to keep it synchronized with the version used in Git plugins.

What are your thoughts on these suggestions? I'm open to any feedback or alternative ideas.

gounthar:
You might want to take a look at #49.
Please do not merge it, let's discuss it here if you don't mind.
Thanks.

jonesbusy:
At the end does it really matter to have same version than the git-client plugin ?

We will not have any classpath issue

I would just use latest jgit version and keep it up-to-date with dependabot like any other dependency

sridamul:
> I would just use latest jgit version and keep it up-to-date with dependabot like any other dependency

This is what I have been thinking of from the beginning, and I don't understand why do we want to have the same version of jgit as in `git-client` or `git` plugin.

gounthar:
I'd like to clarify the rationale behind using the same jgit version as the git-client plugin:

1. Proven Reliability:
   - This version is used on thousands of Jenkins controllers.
   - It undergoes thorough testing by Mark before any upgrades.
   - This approach ensures we're using a reliable jgit version without known regressions.

2. Risk Mitigation:
   - If we were to allow Dependabot to propose the latest available version automatically, we might encounter issues if our tests aren't comprehensive enough.

3. Balanced Approach:
   - While I see benefits in aligning with the git-client plugin version, I'm open to alternatives.
   - We could define the jgit version manually and then use Dependabot to keep it updated.

[...]

PR #49, opened 2024-06-25
Jgit dependency through git client

gounthar:
Please don't merge this PR, this is just a proof of concept, a base to discuss.

After creating #48, I experimented with a potential solution, though I'm not certain if it aligns with Maven best practices.

Here's what I tried:

1. I believe I managed to fetch the `jgit` dependency with the same version as the `git-client` plugin by referencing its `pom.xml`.
   (The project compiles successfully, which suggests it's working)

2. To keep this approach current, we would need to maintain the version number:
   ```xml
   <git.client.version>5.0.0</git.client.version>``
   
Tools like Dependabot or Renovate could help us keep this version up to date.
What are your thoughts on this approach? I'm particularly interested in:
- Whether this aligns with Maven best practices
- Any potential issues or benefits you foresee
- Alternative approaches we might consider

I look forward to hearing your feedback and suggestions.


### Testing done

`mvn clean install`

### Submitter checklist

<!--
Put an `x` into the [ ] to show you have filled the information.
The template comes from https://github.com/jenkinsci/.github/blob/master/.github/pull_request_template.md 
You can override it by creating .github/pull_request_template.md in your own repository 
-->


sridamul:
Hey Bruno, I had tested the changes locally and seems to work fine. IMO Initially, we can stick to this and experiment other approaches like "utilizing updatecli" later.

pom.xml
```diff
@@ -165,6 +166,49 @@
             </plugins>
         </pluginManagement>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <version>3.7.1</version>
```
jonesbusy, line 172: If we go with this solution, I would move the plugin version on the `</pluginManagement>` and execute the plugin only on the core pom

pom.xml
```diff
@@ -165,6 +166,49 @@
...
+                                </artifactItem>
+                            </artifactItems>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>properties-maven-plugin</artifactId>
+                <version>1.2.1</version>
```
jonesbusy, line 197: If we go with this solution, I would move the plugin version on the `</pluginManagement>` and execute the plugin only on the core pom

sridamul:
Hey Bruno, I have inspected a few past PRs from Dependabot to update Jgit dependency in the Git-client-plugin. From my observation, I found that there are only regressions either related to the Jenkins core or the plugin itself. If I am not wrong, then using Dependabot in our tool will not cause any issue (Since, neither core nor plugin involved in our tool). wdyt? (Please correct me if I am wrong) Thanks!

gounthar:
I appreciate your perspective. Upon reflection, I realize that my previous suggestion was based on a gut feeling rather than concrete data.

Given this, I agree that we should proceed with the following approach:

1. Start by implementing Dependabot for our version management.
2. Monitor the results over the next few months.
3. If we encounter any issues, we can reassess our strategy and consider alternatives.

This approach allows us to:
- Begin with a widely-used, automated solution
- Gather real-world data on its effectiveness for our specific use case
- Remain flexible and open to adjustments if needed

Does this sound like a reasonable plan to everyone? I'm open to any thoughts or concerns you might have.

jonesbusy:
Sure fine for me

PR #80, opened 2024-07-15
Pass GH owner from CLI (in case token as access to other organisation

jonesbusy:
In a short future we should also support forking to other organisation.

For example I have the organisation `jonesbusy-automation` and I cannot use it yet for such automation

Also support `GH_TOKEN` which is also used by the gh client https://cli.github.com/manual/gh_help_environment so try to have the same order to lookup for the token

### Testing done

Interactive tests only

### Submitter checklist

### DIFF
--- a/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
+++ b/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
@@ -37,6 +37,9 @@ public class Main implements Runnable {
     @Option(names = {"-r", "--recipes"}, required = true, description = "List of Recipes to be applied.", split = ",", parameterConsumer = CommaSeparatedParameterConsumer.class)
     private List<String> recipes;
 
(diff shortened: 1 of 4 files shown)


gounthar (approved):

sridamul (approved):

plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
```diff
@@ -37,6 +37,9 @@ public static void main(final String[] args) {
     @Option(names = {"-r", "--recipes"}, required = true, description = "List of Recipes to be applied.", split = ",", parameterConsumer = CommaSeparatedParameterConsumer.class)
     private List<String> recipes;
 
+    @Option(names = {"-g", "--github-owner"}, description = "GitHub owner for forked repositories (only username supported for now)")
+    private String githubUsername = Settings.GITHUB_USERNAME;
```
jonesbusy, line 41: To be renamed to githubOwner when we support also organisation.on the `GHService` we should be able to do it transparently if forking to user or an other organisation

PR #86, opened 2024-07-15
New option to parse plugins.txt file

### LINKED ISSUE #77: Allow to pass plugin via a plugin.txt

Allow to pass a list of plugin to modernize via plugins.txt (perhaps we can extract the plugin ID by stripping the version).

User/developer are familiar with the plugins.txt and would allow long list of plugin instead of passing all of them via the CLI

### Upstream changes

_No response_

_No response_

sridamul:
Closes #77 
Added new option to parse plugins.txt file

### Testing done
Written unit tests

### Submitter checklist

### DIFF
--- a/plugin-modernizer-cli/src/test/java/io/jenkins/tools/pluginmodernizer/cli/MainTest.java
+++ b/plugin-modernizer-cli/src/test/java/io/jenkins/tools/pluginmodernizer/cli/MainTest.java
@@ -9,12 +9,14 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 
 import io.jenkins.tools.pluginmodernizer.core.config.Config;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 import picocli.CommandLine;
 
 public class MainTest {
@@ -22,6 +24,9 @@ public class MainTest {
     private CommandLine commandLine;
     private ByteArrayOutputStream outputStream;
     private Main main;
+
+    @TempDir
+    Path tempDir;
 
     @BeforeEach
     public void setUp() {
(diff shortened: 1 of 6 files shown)


plugin-modernizer-core/src/test/resources/plugins.txt
```diff
@@ -0,0 +1,3 @@
+jobcacher-plugin
```
jonesbusy, line 1: This is the repo name, not the plugin name.

I think we are wrongly using pluginName with -plugin suffix but in fact it's the repo name.

Perhaps we should clarifiy this in the tool before this PR.

Passing the plugin short name and getting the SCM url from the plugin center ? Or just ammending the `-plugin` to create the repo ? But I'm not sure it will work 100% of time

jonesbusy (changes requested):
Clarifiy usage of pluginName vs repo name

plugin-modernizer-core/src/test/resources/plugins.txt
gounthar: Better safe than sorry, we should make it super clear in the documentation, the comments, and the code. 👍 
Thanks for catching that early!

sridamul:
I think I misunderstood in the previous conversation. I thought it would easier to look up the repo name instead of plugin name from the UC.

> Passing the plugin short name and getting the SCM url from the plugin center ? Or just ammending the -plugin to create the repo ? But I'm not sure it will work 100% of time

I think it is better to get SCM from uc

sridamul:
Will create a PR to fetch SCM url from UC by passing the plugin short name instead of passing the repo name.

gounthar (approved):

iamrajiv (approved):

plugin-modernizer-core/src/test/resources/plugins.txt
```diff
@@ -0,0 +1,3 @@
+jobcacher
```
jonesbusy, line 1: Did we try with a plugin file that contains version information ? Are we spliting the list correctly ?

gounthar: Here is one I use on a daily basis: https://github.com/jenkins-docs/quickstart-tutorials/blob/main/dockerfiles/plugins.txt .

plugin-modernizer-cli/src/test/java/io/jenkins/tools/pluginmodernizer/cli/MainTest.java
```diff
@@ -56,17 +61,24 @@ public void testGetRecipes() {
...
-    public void testMissingPluginsArgument() {
-        String[] args = {"-r", "recipe1,recipe2"};
+    public void testMissingRecipesArgument() {
+        String[] args = {"-p", "plugin1,plugin2"};
         int exitCode = commandLine.execute(args);
         assertEquals(CommandLine.ExitCode.USAGE, exitCode);
     }
 
     @Test
-    public void testMissingRecipesArgument() {
```
berviantoleo, line 66: I suggest adding a test when the user uses options -f, -r, and -p

jonesbusy (approved):

berviantoleo (approved):

PR #110, opened 2024-07-18
Process Plugins in the Order Defined in plugins.txt

### LINKED ISSUE #105: Process Plugins in the Order Defined in plugins.txt
## Current Behavior
The plugins are currently being processed in an order that appears random, rather than following the order defined in the `plugins.txt` file.

## Observations
- The processing order differs from the order in which plugins are listed in `plugins.txt`.
- The order remains consistent across different machines, suggesting a deterministic but unintended sorting mechanism.

## Desired Outcome
Process the plugins in the exact order they are listed in the `plugins.txt` file.

## Request
Implement a method to ensure plugins are processed in the same sequence as they are defined in the input file.

## Benefits of Implementing This Change
1. Predictable processing order
2. Alignment with user expectations
3. Potential for intentional ordering of plugins based on dependencies or other factors
4. Improved clarity and control for users managing multiple plugins

## Question
Is there a way to modify the current implementation to process the plugins in the order they are defined in `plugins.txt`?

## Next Steps
1. Investigate the current sorting mechanism
2. Develop a solution to preserve the original order from `plugins.txt`
3. Implement and test the new ordering system
4. Update documentation to reflect this change in behavior

Thank you for your consideration.

[plugins.txt](https://github.com/user-attachments/files/16266351/plugins.txt)

sridamul:
Closes #105 
If both `-f` and `-p` options are enabled, then the plugins are processed in the order of plugin file first, followed by the plugins passed through `-p` option.
For Eg: 
plugins.txt
```
git
git-client:1.0.0
jobcacher:2.3.4
```

`java -jar /path/to/jar --plugins login-theme,jobcacher -f path/to/plugin/file --recipes AddPluginsBom -n`
would process in order of git, git-client, jobcacher, login-theme
### Testing done
Unit tests

### Submitter checklist
~- [ ] Link to relevant pull requests, esp. upstream and downstream changes~

### DIFF
--- a/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
+++ b/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
@@ -4,8 +4,9 @@ import java.io.InputStream;
 import java.net.URL;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
@@ -123,7 +124,8 @@ public class Main implements Runnable {
             loadedPlugins.addAll(plugins);
         }
 
-        return new ArrayList<>(new HashSet<>(loadedPlugins));
+        Set<String> uniquePlugins = new LinkedHashSet<>(loadedPlugins);
+        return new ArrayList<>(uniquePlugins);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


gounthar (approved):
I tried it in GitPod with a `plugins.txt` file, and it worked.
Thanks! 🚀

plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/Main.java
```diff
@@ -123,7 +124,8 @@ private List<String> loadPlugins() {
             loadedPlugins.addAll(plugins);
         }
 
-        return new ArrayList<>(new HashSet<>(loadedPlugins));
+        Set<String> uniquePlugins = new LinkedHashSet<>(loadedPlugins);
```
jonesbusy, line 127: More idiomatic way could be

```java
return loadedPlugins.stream().distinct().collect(Collectors.toCollection(ArrayList::new));
```

PR #133, opened 2024-07-30
Cleanup configuration-as-code.version when adding bom

jonesbusy:
Adding bom will remove dependency override which leave an oprhaned property. Cleaning it if needed

I don't see a very valid use case that a plugin will use a different JCasC version than the one provided by the bom

remove_property

### Submitter checklist


gounthar (approved):
Well thought, thanks!

PR #137, opened 2024-07-31
Replace org.json:json by json-api plugin

jonesbusy:
Use existing ReplaceLibrariesWithApiPlugin

We need to think about the TODO and how to get the latest version

The `org.openrewrite.jenkins.AddPluginsBom` cleanup such version override but doesn't change if an old bom is already added on the pom.

I'm thinking to add/implement a cleanup recipe that will just remove the version override if set on the dependency management

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
+++ b/plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
@@ -20,12 +20,29 @@ recipeList:
   - org.openrewrite.jenkins.github.AddTeamToCodeowners
 ---
 type: specs.openrewrite.org/v1beta/recipe
+name: io.jenkins.tools.pluginmodernizer.UpgradeParentVersion
+description: Upgrade the parent version to latest available
+recipeList:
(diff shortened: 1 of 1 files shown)


plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
```diff
@@ -27,3 +27,23 @@ recipeList:
...
+  - io.jenkins.tools.pluginmodernizer.UseJsonApiPlugin
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: io.jenkins.tools.pluginmodernizer.UseJsonApiPlugin
+description: Use JSON API plugin instead of direct dependency
+recipeList:
+  - org.openrewrite.jenkins.ReplaceLibrariesWithApiPlugin:
+      pluginGroupId: io.jenkins.plugins
+      pluginArtifactId: json-api
+      # TODO: version from bom and filtered here ? or managed by renovate ?
```
sridamul, line 45: I think version from BOM would be better to maintain the consistency. wdyt?

sridamul (approved):

PR #140, opened 2024-08-01
Introduce Plugin model and use it where possible and general improvements

jonesbusy:
Some refactoring and code cleanup (specially on the GitHub service and maven invoker)

Use a mutable Plugin object to interact with it using different service created by the PluginModernizer class. Will also simplify testing. Use a more object oriented programing than a collection of utilities and method with parameters. This is far from perfect but will be much easier to understand the code and implement more feature (like parallelism)

This PR also add a skip-push and skip-pull-request parameter for easier testing of recipes without opening noisy pull request.

This also fix some "bug" I found during tests

- If the branch already exists it was not reset to default branch that might causing issue when opening PR 
- When running int dry-mode we will not fork anymore the repo but clone the original repo (from jenkinsci org). The tool also ensure to set the correct remote URL and fetching latest changes when disable dry mode. This will also ensure we don't fork until we are happy with the changes

Also tried to add more javadoc for public methods at least

### Testing done

Automated tests and check full chain with this PR https://github.com/jenkinsci/artifactory-artifact-manager-plugin/pull/60

### Submitter checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,10 @@ Alternatively, you can pass the GitHub owner through the CLI option `-g` or `--g
 
 - `--dry-run` or `-n`: (optional) Enables dry run mode, generating patch files instead of applying changes. The patch files will be generated at `target/rewrite/rewrite.patch` inside the plugin directory if any change is made.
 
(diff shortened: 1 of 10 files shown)


gounthar (approved):
Exactly what we needed, thanks!

Issue #120, opened 2024-07-24
Handling Vastly Outdated Plugins: A Modernization Approach


gounthar:
### Discussed in https://github.com/jenkinsci/plugin-modernizer-tool/discussions/119

<div type='discussions-op-text'>

<sup>Originally posted by **gounthar** July 24, 2024</sup>
## Recent Experience

This week, I attempted to modernize an old plugin still using Java 6. Our current set of recipes failed, necessitating the use of existing OpenRewrite recipes for Java modernization before applying our custom recipes.

## Challenges Encountered

1. **Compilation Verification**: Checking if the plugin still compiled after each set of applied recipes was a major challenge.
2. **JDK Compatibility**: Unable to verify with JDK 11 or 17, I had to revert to JDK 8. For plugins last compiled with JDK 6, verification with JDK 6 might be necessary.

## Proposed Approach

Before applying any recipe, we should:
1. Attempt to compile the plugin (without tests) using the specified Java version.
2. Apply the recipe only if the initial compilation succeeds.

## Considerations

- **Recipe Metadata**: We may lack information about which JDK to use with specific recipes.
- **JDK Version Recipes**: For JDK version-specific recipes, we should run them with the target JDK.

## Tools and Techniques

- Used [sdkman](https://sdkman.io/) to switch between Java versions.
- Successfully upgraded a plugin to JDK 17. 🥳

## Potential Solutions

1. **Script Generation**: Create scripts to switch JDK versions based on recipes using Temurin JDK versions.
2. **Java-centric Approach**: Explore more Java-native solutions instead of using `exec()`.
3. **Jenkins Controller Delegation**: Use API calls to delegate compilation checks to another Jenkins controller.
4. **GitHub Actions**: Utilize GitHub Actions on our forked plugin repositories for compilation checks.

## Questions for Consideration

1. Can we integrate JDK switching directly into our product?
2. Is there a more Java-centric approach to handle compilation checks?
3. Should we explore delegating checks to external systems (Jenkins controllers or GitHub Actions)?
4. Are there alternative approaches we haven't considered?

Your feedback and suggestions on these ideas would be greatly appreciated.</div>

PR #147, opened 2024-08-03
Cache missing JDK and Add migration recipes

sridamul:
Related to Issue #120 Discussion #119

Changes made:
- Caches JDK by downloading and extracting temurin binaries from [GitHub](https://github.com/orgs/adoptium/repositories?q=temurin+binaries) based on OS, if the required JDK is not present in the system. (maybe we can also add options to get JDK paths, if it is present in different location?)
- Added migration recipe to upgrade java to 17 to check the viability of new changes (will add more migration recipes later)

Tested the new recipe with some plugins. Not all plugins are successfully migrated, need to cover the edge cases too.

What can we do now?
Now the tool can compile plugins with older JDK versions, if the JDK not present, then it caches it in `{user.dir}/.jdks/` (directory used by IntelliJ IDEA to store temurin JDKs), followed by executing the `runOpenRewrite()` fn with JDK 17.

What else?
Right now I hard coded the java version in the `getJdkPath()`, need to figure out a way to get opt JDK version to compile each plugin. 

Other suggestions and improvements are welcomed

### Testing done
Interactive testing only.
Manually tested the caching process in Windows and Linux environment.

### Submitter checklist
~- [ ] Link to relevant pull requests, esp. upstream and downstream changes~

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
@@ -5,6 +5,7 @@ import io.jenkins.tools.pluginmodernizer.core.config.Config;
 import io.jenkins.tools.pluginmodernizer.core.config.Settings;
 import io.jenkins.tools.pluginmodernizer.core.github.GHService;
 import io.jenkins.tools.pluginmodernizer.core.model.Plugin;
+import io.jenkins.tools.pluginmodernizer.core.utils.JdkFetcher;
 import io.jenkins.tools.pluginmodernizer.core.utils.JenkinsPluginInfo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -57,6 +58,8 @@ public class PluginModernizer {
     private void process(Plugin plugin) {
         try {
 
+            JdkFetcher jdkFetcher = new JdkFetcher();
+
             // Determine repo name
             plugin.withRepositoryName(JenkinsPluginInfo.extractRepoName(
                     plugin.getName(), config.getCachePath(), config.getJenkinsUpdateCenter()));
@@ -70,9 +73,10 @@ public class PluginModernizer {
                 plugin.removeLocalData();
             }
             plugin.fetch(ghService);
-            plugin.clean(mavenInvoker);
+            plugin.compile(mavenInvoker, jdkFetcher.getJdkPath("8"));
             plugin.checkoutBranch(ghService);
-            plugin.runOpenRewrite(mavenInvoker);
+            plugin.runOpenRewrite(mavenInvoker, jdkFetcher.getJdkPath("17"));
+            plugin.verify(mavenInvoker, jdkFetcher.getJdkPath("17"));
             plugin.commit(ghService);
             plugin.push(ghService);
             plugin.openPullRequest(ghService);
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
@@ -72,10 +72,12 @@ public class PluginModernizer {
                 plugin.removeLocalData();
             }
             plugin.fetch(ghService);
-            plugin.compile(mavenInvoker, jdkFetcher.getJdkPath("8"));
+            plugin.withJdkPath(jdkFetcher.getJdkPath("8"));
+            plugin.compile(mavenInvoker);
             plugin.checkoutBranch(ghService);
-            plugin.runOpenRewrite(mavenInvoker, jdkFetcher.getJdkPath("17"));
-            plugin.verify(mavenInvoker, jdkFetcher.getJdkPath("17"));
+            plugin.withJdkPath(jdkFetcher.getJdkPath("17"));
+            plugin.runOpenRewrite(mavenInvoker);
+            plugin.verify(mavenInvoker);
             plugin.commit(ghService);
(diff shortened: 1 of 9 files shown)


gounthar (approved):

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/utils/JdkFetcher.java
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Downloads the JDK for the specified version from the appropriate URL and saves it to the local directory.
+     *
+     * @param jdkVersion The version of the JDK (e.g., "8").
+     * @return The path to the downloaded JDK file.
+     * @throws IOException          If an I/O error occurs.
+     * @throws InterruptedException If the operation is interrupted.
+     */
+    @SuppressFBWarnings(value = "URLCONNECTION_SSRF_FD", justification = "false positive")
+    private Path downloadJdk(String jdkVersion) throws IOException, InterruptedException {
```
jonesbusy, line 85: What do you think about an Enum ? We pass always the major version, never the minor version or path.

 Adding a new JDK could be just adding a new value on the enum

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/utils/JdkFetcher.java
```diff
@@ -0,0 +1,270 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * A utility class for fetching the latest JDK releases from the Adoptium GitHub repository.
+ */
+@SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "false, positive")
+public class JdkFetcher {
+    private static final Logger LOG = LoggerFactory.getLogger(JdkFetcher.class);
+
+    private static final String GITHUB_API_URL = "https://api.github.com/repos/adoptium";
```
jonesbusy, line 32: This must be on settings. An perhaps an improvement to change the API URL

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/utils/JdkFetcher.java
```diff
@@ -0,0 +1,270 @@
...
+     * Gets the path to the JDK directory for the specified JDK version. If the JDK is not already downloaded,
+     * it triggers the download and setup process.
+     *
+     * @param jdkVersion The version of the JDK (e.g., "8").
+     * @return The path to the JDK directory.
+     * @throws IOException          If an I/O error occurs.
+     * @throws InterruptedException If the operation is interrupted.
+     */
+    public Path getJdkPath(String jdkVersion) throws IOException, InterruptedException {
+        Path jdkPath = Paths.get(DOWNLOAD_DIR, "plugin-modernizer-jdk-" + jdkVersion);
```
jonesbusy, line 45: Why we don't rely on CacheManager for those kind things ?

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/utils/JdkFetcher.java
```diff
@@ -0,0 +1,270 @@
...
+
+/**
+ * A utility class for fetching the latest JDK releases from the Adoptium GitHub repository.
+ */
+@SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "false, positive")
+public class JdkFetcher {
+    private static final Logger LOG = LoggerFactory.getLogger(JdkFetcher.class);
+
+    private static final String GITHUB_API_URL = "https://api.github.com/repos/adoptium";
+    private static final String DOWNLOAD_DIR = System.getProperty("user.home") + "/.jdks/";
```
jonesbusy, line 33: What if the user.home is not writable ? Why we don't rely on cache dir to specify an writable location

pom.xml
```diff
@@ -46,6 +46,7 @@
     <openrewrite.bom.version>2.16.0</openrewrite.bom.version>
     <openrewrite.maven.plugin.version>5.37.0</openrewrite.maven.plugin.version>
     <openrewrite.jenkins.version>0.9.0</openrewrite.jenkins.version>
+    <openrewrite.rewrite-java.version>2.21.0</openrewrite.rewrite-java.version>
```
jonesbusy, line 49: Isn't it part of the bom ?

jonesbusy (approved):
Thanks, some minor design issue on my point of view. Feel free to address them on this PR or later

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
```diff
@@ -70,9 +73,10 @@ private void process(Plugin plugin) {
                 plugin.removeLocalData();
             }
             plugin.fetch(ghService);
-            plugin.clean(mavenInvoker);
+            plugin.compile(mavenInvoker, jdkFetcher.getJdkPath("8"));
```
jonesbusy, line 76: Will cause issue on plugin that will require a core version or dependencies requiring modern JDK (or even 11)

sridamul: Yep! Need to figure a way to choose JDK automatically (based on plugins) instead of hard-coding it.

jonesbusy:
Thanks. Let me know when you are ready to merge this PR

iamrajiv (approved):

sridamul:
Hey @jonesbusy, I think this PR is ready to merge, we can use enums when resolving #153. 
Thanks!

PR #157, opened 2024-08-05
Add tests for MavenInvoker

sridamul:
Added a test plugin to test the Maven Invoker functions.
If there is better way to test MavenInvoker fn, let me know. Thanks!

### Submitter checklist


jonesbusy:
I wouldn't call maven directly from such unit tests

What about mocking `Invoker invoker = new DefaultInvoker()` perhaps moving it to a private member instead of local variable ?

Even if it's private it can be mocked (See https://github.com/jenkinsci/plugin-modernizer-tool/blob/main/plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/github/GHServiceTest.java#L72)

So the `invoker.execute(request);` can be mocked and asserting the correct request (maven are are passed)

jonesbusy (changes requested):
Try mocking the `Invoker` class (move it as a member rather than local variable)

jonesbusy:
Not sure if planning to work on this, but since we added Guice unit testing the MavenInvoker is probably more easy now

PR #158, opened 2024-08-05
Use SDK man java if possible

jonesbusy:
For me JDK download doesn't work on Linux. Seems the permission of binaries are lost and java cannot be executed

See https://cdeliveryfdn.slack.com/archives/C071YTZ807N/p1722868913559519

I'm also using sdkman to install my JDKs, so I can workarround this issue by simply using them

### Submitter checklist


jonesbusy:
Will fix tomorrow

gounthar (commented):
There is already a bash file that helps with installing sdkman and jdk from 8 to 21 thanks to it.
The available versions are already kept up to date thanks to an updatecli manifest.
I could also update this property file the same way with updatecli.

jonesbusy:
Ready for review. I've updated the updatecli configuration that was using the forcecraete at wrong location. And use the search pattern for file to make it more generic and working on this PR

gounthar (approved):

PR #159, opened 2024-08-06
Abort plugin modernization on errors and set target JVM

jonesbusy:
Errors were missing in some case and when maven return a non-zero status code.

Also ensure to abort the process when such error occured 

Set also the target JDK configurable to test compiling plugin with Java 21 until we have JVM auto-detection for plugin source

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenInvoker.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenInvoker.java
@@ -150,6 +150,7 @@ public class MavenInvoker {
             handleInvocationResult(plugin, result);
         } catch (MavenInvocationException e) {
             LOG.error(plugin.getMarker(), "Maven invocation failed: ", e);
+            plugin.addError(e);
         }
     }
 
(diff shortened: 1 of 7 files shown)


plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenInvoker.java
```diff
@@ -150,6 +150,7 @@ private void invokeGoals(Plugin plugin, String... goals) {
             handleInvocationResult(plugin, result);
         } catch (MavenInvocationException e) {
             LOG.error(plugin.getMarker(), "Maven invocation failed: ", e);
+            plugin.addError(e);
```
jonesbusy, line 153: Was silently failing :(

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/MavenInvoker.java
```diff
@@ -223,6 +224,8 @@ private void handleInvocationResult(Plugin plugin, InvocationResult result) {
             if (result.getExecutionException() != null) {
                 LOG.error(plugin.getMarker(), "Execution exception occurred: ", result.getExecutionException());
                 plugin.addError(result.getExecutionException());
+            } else {
+                plugin.addError(new MavenInvocationException("Build failed with code: " + result.getExitCode()));
```
jonesbusy, line 228: In case the maven command finish but return a non-zero status code in case of build failure. Was ignored

gounthar (approved):

PR #161, opened 2024-08-07
Add executable permissions for jdk binaries

sridamul:
Resolve the executable permissions after extracting the JDK

see https://cdeliveryfdn.slack.com/archives/C071YTZ807N/p1722868913559519

### Testing done
Interactive testing with Ubuntu 24.04

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/utils/JdkFetcher.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/utils/JdkFetcher.java
@@ -16,6 +16,10 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
+import java.util.Set;
+import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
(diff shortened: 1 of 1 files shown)


jonesbusy:
Instead of fixing the permissions afterward there is no way to keep the permissions when untaring the archive ?

sridamul:
> Instead of fixing the permissions afterward there is no way to keep the permissions when untaring the archive ?

By default, [TarArchiveEntry](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/tar/TarArchiveEntry.html) class uses `644` permission for [files](https://github.com/apache/commons-compress/blob/35adf67d67a0d6bc323c9a7e97a78da1d0625cab/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java#L204) and `755` for [directories,](https://github.com/apache/commons-compress/blob/35adf67d67a0d6bc323c9a7e97a78da1d0625cab/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java#L201) that's the reason not to have `+x` for bin files. From my understanding, we can use [setMode()](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/tar/TarArchiveEntry.html#setMode(int)) and [getMode()](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/tar/TarArchiveEntry.html#getMode()) to preserve the file permissions for each entry, which then can be applied via `setPosixFilePermissions()` fn. I think, which might be equivalent to directly applying the executable permissions for the bin files. 
Please correct me If i am wrong. 
Thanks!

jonesbusy:
Looks of for me. Seems there is not better way.

I found this code from jenkins core https://github.com/jenkinsci/jenkins/blob/c626519b89bcee0840a6d658b06cd79808d41294/core/src/main/java/hudson/tools/ZipExtractionInstaller.java#L146

That basically set the file permissions after extraction

jonesbusy (approved):

PR #168, opened 2024-08-09
Remove pointless message for slf4j

jonesbusy:
Seems recent upgrade of slf4j (Maybe https://github.com/jenkinsci/plugin-modernizer-tool/pull/164) started to print message when tool is starting

```
SLF4J(I): Connected with provider of type [ch.qos.logback.classic.spi.LogbackServiceProvider]
```

This is not useful at all, we know that we are using logback as logger implementation

Removing this message 

See https://stackoverflow.com/questions/78063648/how-to-stop-slf4j-printing-a-pointless-information-message-at-startup

### Testing done

Ensuring the message disapear when running the tool after setting the system property

### Submitter checklist


jonesbusy:
See also https://github.com/qos-ch/slf4j/issues/422

sridamul (approved):

PR #187, opened 2024-08-18
Verify with jdk that used in compile phase

sridamul:
Right now, the plugin is verified with `JDK 17` even though a plugin with JDK 8 or JDK 11 supported is compiled.
This PR verifies with JDK as used in compile phase.

### Submitter checklist


jonesbusy:
I think this will not work if we fully refresh a plugin. It will not be compatible with Java 8 but only 11 and even soon Java 17 due to recent core minimum requirement. And a number of plugin that will bundle dependencies requiring Java 17

I think we

- EIther rely on the metadata (but we miss some logic related to minimum JDK build)
- Use the same logic but downgrading the verify until we find a correct one

Probably the 2nd is more quick to implement for now

- Compile with max JDK
- If it fail downgrade it until it works

jonesbusy (approved):

PR #195, opened 2024-08-18
Add JSON Utils and CacheEntry

jonesbusy:
Add some JSON utility using GSON implementation and `CacheEntry` abstract class

There are some other place we need to uniformize the serializaton that still using Jackson and not this class

Also the `UpdateCenterData` need to contains field  we need instead of storing a reference on `JsonNode`. Need to be done in an other PR to manage our cached entities the same way (maybe using `@Expose`)

### Submitter checklist


jonesbusy:
I will close this PR, I have more content to add

PR #212, opened 2024-08-22
Fetch metadata from cache to use JDK versioned computed

jonesbusy:
Fetch metadata from cache to use JDK versioned computed ([PR still](https://github.com/jenkinsci/plugin-modernizer-tool/pull/201))

### Testing done

Tested to run 2 time fetch metadata. The second execution is almost instant because it used the data from the local cache

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
@@ -101,9 +101,13 @@ public class PluginModernizer {
                 LOG.info("Plugin {} has errors. Will not process this plugin.", plugin.getName());
             }
 
+            // Set the metadata from cache if available
+            plugin.setMetadata(cacheManager.get(
+                    Path.of(plugin.getName()), CacheManager.PLUGIN_METADATA_CACHE_KEY, PluginMetadata.class));
+
             // Compile the plugin with the first JDK that compile it
-            JDK jdkCompile = compilePlugin(plugin);
             if (!config.isFetchMetadataOnly()) {
+                JDK jdkCompile = compilePlugin(plugin);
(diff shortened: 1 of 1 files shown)


plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
```diff
@@ -117,21 +121,25 @@ private void process(Plugin plugin) {
...
-            LOG.debug(
-                    "Moved plugin {} metadata to cache: {}",
-                    plugin.getName(),
-                    plugin.getMetadata().getLocation().toAbsolutePath());
+            // Collect metadata and move metadata from the target directory of the plugin to the common cache
+            if (plugin.getMetadata() == null) {
+
+                plugin.collectMetadata(mavenInvoker);
+
+                CacheManager pluginCacheManager = new CacheManager(Path.of(Settings.TEST_PLUGINS_DIRECTORY)
```
jonesbusy, line 129: I'm not happy with this code. This probably need to be moved somewere (even a private method would fix). Will clean on an other PR

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/impl/PluginModernizer.java
```diff
@@ -101,9 +101,13 @@ private void process(Plugin plugin) {
                 LOG.info("Plugin {} has errors. Will not process this plugin.", plugin.getName());
             }
 
+            // Set the metadata from cache if available
+            plugin.setMetadata(cacheManager.get(
```
jonesbusy, line 105: Maybe deserve `plugin.loadMetadata`

gounthar (approved):
Thanks for this very encouraging first step in dealing with metadata cache. 👍

PR #217, opened 2024-08-23
Add static preconditions for pom.xml

jonesbusy:
An improvement would be to store metadata with those failing details. So that we can identify plugin that require (manual ?) fix because they are so outdated

At least now the error is more meaningful

### Testing done

Didn't write test for now, but did some interfactive tests

error2
error

### Submitter checklist


gounthar (approved):
Good idea, thanks! 👍

jonesbusy:
Related to https://github.com/jenkinsci/plugin-modernizer-tool/issues/215

Doesn't fix, but make error clearer.

We can improve to build partial metadata by only static pom parsing

PR #223, opened 2024-08-24
Add PR template using JTE

jonesbusy:
I'm not 100% satisfied but is better than before 

Example https://github.com/jenkinsci/gson-api-plugin/pull/39

Last time I did templating on Java (10 years ago) I was using Velocity but it seems there are much better alternatives in the more modern world. JTE looked a good and easy engine

Templates are rendered and logged in DEBUG mode.

One future improvement is to adapt the title depending on the recipes (like using tags)

Since we are only relying on recipes from the plugin modernizer tool I've enforced via a test that all field (description, display name etc) are set to simplify remplates

### Testing done

https://github.com/jenkinsci/gson-api-plugin/pull/39

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/pom.xml
+++ b/plugin-modernizer-core/pom.xml
@@ -25,6 +25,10 @@
       <artifactId>gson</artifactId>
     </dependency>
     <dependency>
+      <groupId>gg.jte</groupId>
+      <artifactId>jte</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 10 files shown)


gounthar (approved):
That's a fantastic step in the direction we are aiming for. 🥳
Thanks! 🙏

sridamul (approved):
Thanks!

PR #262, opened 2024-09-13
Ensure to push updated refs

jonesbusy:
Seens on some run. In some case the local git repository is outdated even if the fork is sync. Previously we reseted only the unstages change via HARD reset.

Now we reset it to the default origin branch.

When pushing we need to force in case the branch is outdated or push will fail due to previous reset

### Submitter checklist


gounthar (approved):
Nice, thanks! 🙏

Issue #240, opened 2024-08-30
Distribute the CLI as a container


jonesbusy:
### What feature do you want to see added?

Support any host with a container runtime (Docker, Kubernetes,....)

- GithubPackages ? 
- Jenkins4eval org https://hub.docker.com/u/jenkins4eval ?

gounthar:
Fixed by #269 .

PR #269, opened 2024-09-23
Distribute the CLI as a container

gounthar:
This Pull Request aims to address #240.

The docker image is built and then made available as a package in the repository itself.

It currently works on my fork.

I modified the `Dockerfile` to enable the use of the local Maven cache during both GitHub Actions builds and local builds.

Additionally, I added scripts that allow the local Maven cache to be used when building the Docker image locally.

### Testing

```bash
docker run -e GH_TOKEN=${GH_TOKEN} -e GH_OWNER=${GH_OWNER} \
-v $(pwd)/plugins.txt:/plugins.txt \
ghcr.io/gounthar/plugin-modernizer-tool:ghcr \
--plugin-file /plugins.txt --recipes FetchMetadata
```

### Submitter checklist

### DIFF
--- a/.github/workflows/github-docker-registry-push.yml
+++ b/.github/workflows/github-docker-registry-push.yml
@@ -12,6 +12,7 @@ on:
   workflow_dispatch: # Allow the workflow to be triggered manually.
 
 env:
+  DOCKER_BUILDKIT: 1 # Enable Docker BuildKit for faster builds.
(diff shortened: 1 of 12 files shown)


gounthar:
We don't need to mount the maven dependencies, it's there to speed things up.
The image would build without it, for sure.
It just allows us to use the cache mechanism available in GitHub so each docker image build doesn't have to retrieve all dependencies every time.

Issue #215, opened 2024-08-22
'Build failed' when fetching metadata for some plugins.


gounthar:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
  Paste the output here
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

WSL2

### Reproduction steps

I have a `plugins.txt` file which lists tons of plugins.
`job-poll-action-plugin` is part of that list.
`java -jar plugin-modernizer-cli/target/jenkins-plugin-modernizer-999999-SNAPSHOT.jar --plugin-file plugins.txt --recipes FetchMetadata`

I don't have any log available in the `logs` directory.
If I launch directly for this plugin, not using a list, I get an error in the logs.
```
java -jar plugin-modernizer-cli/target/jenkins-plugin-modernizer-999999-SNAPSHOT.jar --plugins vagrant --recipes FetchMetadata
Starting Plugin Modernizer
Skipping fork for plugin vagrant as only metadata is required
Skipping sync for plugin vagrant as only metadata is required
Fetching plugin code locally vagrant...
Plugin processing error. Check the logs at logs/vagrant.log
*************
Plugin: vagrant
Error: Plugin processing error. Check the logs at logs/vagrant.log
*************
Plugin Modernizer aborted.
```
The logs are:
```
2024-08-22T12:09:16.044Z [INFO] [Thread=StreamPumper-systemOut] - i.j.t.p.core.impl.MavenInvoker # [INFO] Scanning for projects... 
2024-08-22T12:09:18.653Z [INFO] [Thread=StreamPumper-systemOut] - i.j.t.p.core.impl.MavenInvoker # [WARNING] The requested profile "consume-incrementals" could not be activated because it does not exist. 
2024-08-22T12:09:18.654Z [INFO] [Thread=StreamPumper-systemOut] - i.j.t.p.core.impl.MavenInvoker # [WARNING] The requested profile "might-produce-incrementals" could not be activated because it does not exist. 
2024-08-22T12:09:18.855Z [INFO] [Thread=StreamPumper-systemOut] - i.j.t.p.core.impl.MavenInvoker # [INFO]  
2024-08-22T12:09:18.857Z [INFO] [Thread=StreamPumper-systemOut] - i.j.t.p.core.impl.MavenInvoker # [INFO] -------------------< org.jenkins-ci.plugins:vagrant >------------------- 
[...]

PR #287, opened 2024-10-02
chore(maven): Add Jenkins and BOM Versions to Properties

gounthar:
## Purpose

We need to include Jenkins and BOM (Bill of Materials) versions for outdated plugins that are still built with JDK versions older than JDK 8. This change will allow Maven 3.9.x to consider applying recipes by force-modifying their `pom.xml` files. These modifications are based on the "[Improve a plugin tutorial](https://www.jenkins.io/doc/developer/tutorial-improve/)".

## Implementation Details

1. The mechanism to keep these properties up-to-date using `updatecli` is ready but will be implemented in a separate PR.
2. A brute-force mechanism to modify the `pom.xml` is also prepared, but this will require further discussion before implementation.

This change is related to issue #215.

## Future Steps

1. Implement `updatecli` mechanism for property updates (separate PR)
2. Discuss and potentially implement the brute-force `pom.xml` modification mechanism

## Testing Performed

No extensive testing was performed as no code was changed. However, a `mvn clean install` was run to ensure build integrity.

### Submitter checklist
- ~~Ensure you have provided tests - that demonstrates feature works or fixes the issue~~


plugin-modernizer-core/src/main/resources/versions.properties
```diff
@@ -1 +1,5 @@
 openrewrite.maven.plugin.version = ${openrewrite.maven.plugin.version}
+jenkins.version = 2.440.2
```
jonesbusy, line 2: Why not 2.440.3 ? 

Perhaps I would be more specific on the property name ? Like jenkins.minimum.version

gounthar: You're right, it was a lame attempt at getting my `updatecli` script to make its first pull request afterward. 😊 
I'll correct it.

jonesbusy (approved):

PR #306, opened 2024-10-07
Increment recommanded version

jonesbusy:
Increment recommanded core version

The BOM `2.440` is not receiving updates. Last one is `3435.v238d66a_043fb_` (https://github.com/jenkinsci/bom?tab=readme-ov-file#depending-on-older-versions)

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
+++ b/plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
@@ -133,7 +133,7 @@ recipeList:
     # https://github.com/jenkins-infra/jenkins-version?
   - org.openrewrite.jenkins.UpgradeVersionProperty:
       key: jenkins.version
-      minimumVersion: 2.440.3
+      minimumVersion: 2.452.4
   - io.jenkins.tools.pluginmodernizer.UpgradeParentVersion
   - io.jenkins.tools.pluginmodernizer.RemoveDependencyVersionOverride
   - io.jenkins.tools.pluginmodernizer.RemoveExtraMavenProperties


plugin-modernizer-core/src/main/resources/versions.properties
```diff
@@ -1,5 +1,5 @@
 openrewrite.maven.plugin.version = ${openrewrite.maven.plugin.version}
-jenkins.minimum.version = 2.440.3
-jenkins.plugin.parent.version = 4.80
-bom.version = 3413.v0d896b_76a_30d
-bom.base = bom-2.440.x
+jenkins.minimum.version = 2.452.4
+jenkins.plugin.parent.version = 4.88
+bom.version = 3435.v238d66a_043fb_
+bom.base = bom-2.452.x
```
gounthar, line 5: # Clarification on Property Usage for Legacy Plugin Upgrades

I apologize for not creating a PR to demonstrate the use of these properties earlier. Upon reflection, I realize the property names should have been more descriptive.

## Purpose of the Properties

The intended purpose of these properties is to facilitate a basic upgrade of the `pom.xml` file for projects still using JDK versions older than 8. 

## Correction on Version Management

Initially, I had implemented `updatecli` manifests to keep these versions current. However, I now recognize this approach is counterproductive. These versions should remain static to maintain compatibility with older JDKs.

## Current Implementation

In [my working branch](https://github.com/gounthar/plugin-modernizer-tool/blob/add-ignore-jdk7-error/plugin-modernizer-core/src/main/resources/versions.properties), I've set `jenkins.plugin.parent.version` to `4.51`. This version is significant because it's [the last version](https://github.com/jenkinsci/plugin-pom) compatible with Java 8.

> Note: Since version 4.52, the plugin parent POM requires Jenkins 2.361 or newer and JDK 11 or newer.

## Proposed Changes

To improve clarity, we should consider renaming these properties to indicate that they should remain unchanged and not be updated. This would help prevent confusion and maintain the intended functionality for legacy plugin support.

Your thoughts on appropriate naming conventions for these static properties would be appreciated.

jonesbusy: Thanks for the detail. Until we find better name I will just change the recipe.yml

In anycase I'm blocked for running new recommended upgrade because of https://github.com/openrewrite/rewrite/pull/4404

jonesbusy: What about something like this (with a comment)

```properties
# Remediation properties are used to upgrade legacy plugins (pre-Java 8) to a minimal working version
# DO NOT changes those values
remediation.jenkins.minimum.version = 2.440.3
remediation.jenkins.plugin.parent.version = 4.80
remediation.bom.version = 3413.v0d896b_76a_30d
remediation.bom.base = bom-2.440.x

```

gounthar: Fine with me, thanks 👍

PR #310, opened 2024-10-09
Add (empty) function to remediate a precondition error of a plugin

jonesbusy:
Relate to https://github.com/jenkinsci/plugin-modernizer-tool/pull/307/files

But without using any transformation

@gounthar Perpaps you can "cherry-pick" code of the `PomModifier` using of each precondition error remediate function ?

For the moment we don't have tests for the `PluginModernizer` class but I've tested the function is called with the `vagrant` plugin for example. Right now all function return false because nothing is done. I left TODOs in the code

remediation

### Submitter checklist


gounthar (approved):
Thanks, I will try to base my work on this. 👍

PR #317, opened 2024-10-10
Improve and add more metadata flags

jonesbusy:
### Testing done

Automated tests.

Tested with some plugin like asm-api or vagrant. Check the metadata is correctly updated and persisted

One of the reason of this PR is that I would like to setup some conditional on recipes.

Right now the `recipes.yml` is just a flat collection of recipes that openrewrite can run.

But we have no way to setup conditional based on metadata. Examples

- Adding bom should not apply if there is no plugin dependencies
- Adding security scan should not be added or removed if the plugin doesn't contains any source file (like most API plugin). Running this scan is a waste of time and resource
- We might want to run renovate/dependabot more often for API plugins than standard plugin (to ensure to test bom as early as possible)

etc...

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/PluginMetadata.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/PluginMetadata.java
@@ -8,6 +8,7 @@ import io.jenkins.tools.pluginmodernizer.core.model.PreconditionError;
 import java.io.Serializable;
 import java.nio.file.Path;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/PluginMetadata.java
```diff
@@ -112,11 +113,21 @@ public void setFlags(Set<MetadataFlag> flags) {
...
+        flags.add(flag);
+    }
+
     public boolean hasFlag(MetadataFlag flag) {
-        return flags.contains(flag);
+        return flags != null && flags.contains(flag);
     }
 
     public Set<PreconditionError> getErrors() {
+        if (errors == null) {
```
jonesbusy, line 128: @gounthar I saw the same nullpointer this morning

gounthar (approved):

gounthar:
Looks good to me, thanks!

PR #413, opened 2024-12-07
Fix typos in JReleaser workflow

jonesbusy:
Tested on my fork this time

### DIFF
--- a/jreleaser.yml
+++ b/jreleaser.yml
@@ -16,7 +16,6 @@ project:
 release:
   github:
     enabled: true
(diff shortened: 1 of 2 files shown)


jreleaser.yml
```diff
@@ -16,7 +16,6 @@ project:
 release:
   github:
     enabled: true
-    owner: jenkins-infra
```
jonesbusy, line 19: Not needed, current owner is used

gounthar (approved):

PR #416, opened 2024-12-08
Refresh stats URL

jonesbusy:
Refresh stats URL

Perhaps could be automated using updatecli ?


gounthar:
That should be doable with updatecli.

PR #417, opened 2024-12-08
Rename parent-pom to pom

jonesbusy:
No need to specify parent, pom is enough

And wondering if magically it fixes permission issue

https://github.com/jenkins-infra/repository-permissions-updater/pull/4127/files#diff-fc4047510f5e367e5511180a4da22064a9664ba4e829e36604125741cccb2bcb

I'm wondering if `*` would match `pom` but not `parent-pom` due to the dash ?


jonesbusy:
Still 403. Was not the issue

PR #418, opened 2024-12-08
Include build date on version

jonesbusy:
Useful information when we distribute it

Format is now

```
plugin modernizer 888.v02298b_910723 (2024-12-08T09:23:53Z)
```


gounthar (approved):
Great, thanks 👍

Issue #370, opened 2024-11-06
Pull Requests: Only List Recipes with Repository Impact


gounthar:
### What feature do you want to see added?

Currently, our process involves listing all applied recipes in the pull request title, regardless of whether they resulted in changes to the repository.

For example, [this PR](https://github.com/jenkinsci/configurationslicing-plugin/pull/295) was renamed by a maintainer because the original title was misleading compared to the actual (minor) changes.

To improve clarity, we should:

1. Only include recipe names in the pull request title if they resulted in changes to the repository.
2. Consider adding a new section in the pull request body to list recipes that did not have an effect, if desired.

This approach will ensure the pull request titles accurately reflect the substantive changes being introduced, making it easier for reviewers to understand the impact.

jonesbusy:
Not relevant since https://github.com/jenkins-infra/plugin-modernizer-tool/pull/419

PR #419, opened 2024-12-08
Run only one recipe per execution

jonesbusy:
This simplify at lot and will allow to work on new feature like user friendly PR title or discard https://github.com/jenkins-infra/plugin-modernizer-tool/issues/370

This PR also fix one bug when a repo was containing unversionned files (like a rewrite.yml) causing build issue.

A `git clean` is now done after a `git --reset hard`

I will also take a not about the collection of the metadata after the modernization. This should probably done before opening the PR (and not after). Or maybe we can just patch the PR title


gounthar (approved):
Good idea, thanks! 🤗

PR #431, opened 2024-12-10
Add gson-api replacement

jonesbusy:


gounthar (approved):

gounthar:
I'll try to do something with `updatecli`.

jonesbusy:
> I'll try to do something with `updatecli`.

Thanks. Or perhaps the recipe need a fix upstream to have an optional version

Probably all of the API plugin are on the bom. And the bom should be added as long we have a plugin dependency

Issue #435, opened 2024-12-10
Reduce fat jar size


jonesbusy:
### What feature do you want to see added?

The far JAR package everything which create an uber jar of 100Mib

Let's see if

- We can remove transitive dependencies
- Optimize the jar by shading instead of packaging uber jar
- Any other option

PR #445, opened 2024-12-11
Optimize far jar

jonesbusy:
This is partial implementation of https://github.com/jenkins-infra/plugin-modernizer-tool/issues/435

This reduce from 180Mib to 50Mib

I'm sure we can easily reduce it to 50% or much more

I identifiy some dependency we must break 

CLI --> openrewrite-core (The `Recipe` class) Like here https://github.com/jenkins-infra/plugin-modernizer-tool/blob/main/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/converter/RecipeConverter.java#L4
CLI --> openrewrite-xml (The `org.openrewrite.xml.tree.Xml` class) like here https://github.com/jenkins-infra/plugin-modernizer-tool/blob/main/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/MetadataFlag.java#L8

### DIFF
--- a/plugin-modernizer-core/pom.xml
+++ b/plugin-modernizer-core/pom.xml
@@ -133,6 +133,12 @@
     <dependency>
       <groupId>org.openrewrite.recipe</groupId>
       <artifactId>rewrite-testing-frameworks</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.testcontainers</groupId>
+          <artifactId>testcontainers</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/plugin-modernizer-core/pom.xml
+++ b/plugin-modernizer-core/pom.xml
@@ -139,11 +139,6 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


plugin-modernizer-core/pom.xml
```diff
@@ -159,6 +160,12 @@
       <artifactId>rewrite-test</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>uk.org.webcompere</groupId>
+      <artifactId>system-stubs-jupiter</artifactId>
+      <version>2.1.7</version>
+      <scope>test</scope>
```
jonesbusy, line 167: Scope test was missing

pom.xml
```diff
@@ -282,6 +278,11 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-junit-jupiter</artifactId>
+      <scope>test</scope>
```
jonesbusy, line 284: Scope test was missing

gounthar (approved):
Thanks, that looks like a fantastic start!

PR #451, opened 2024-12-13
Add Maven Invoker Integration Test and Ensure Successful Build

SohamJuneja:
Added a new integration test MavenInvokerIntegrationTest to validate Maven commands (clean and verify) using the Maven Invoker API.
Ensured the integration test runs successfully and verifies the build process.
Added the maven-invoker dependency to the plugin-modernizer-cli module for integration testing.
Updated the pom.xml in plugin-modernizer-cli to include required configurations for running tests.
Verified the project builds successfully with all tests passing.

### DIFF
--- a/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/command/ValidateCommand.java
+++ b/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/command/ValidateCommand.java
@@ -36,6 +36,8 @@ public class ValidateCommand implements ICommand {
             modernizer.validate();
             LOG.info("GitHub owner: {}", modernizer.getGithubOwner());
             LOG.info("Maven home: {}", config.getMavenHome());
+            LOG.info("Maven version: {}", modernizer.getMavenVersion());
+            LOG.info("Java version: {}", modernizer.getJavaVersion());
             LOG.info("Cache path: {}", config.getCachePath());
         } catch (ModernizerException e) {
             LOG.error("Validation error");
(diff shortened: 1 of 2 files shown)


SohamJuneja:
I noticed that the CI pipeline is failing due to the MavenInvokerIntegrationTest. I am investigating the root cause and will push updates shortly. The test works locally, but it seems to encounter issues in the CI environment

plugin-modernizer-cli/src/test/java/io/jenkins/tools/pluginmodernizer/cli/MavenInvokerIntegrationTest.java
```diff
@@ -0,0 +1,22 @@
+package io.jenkins.tools.pluginmodernizer.cli;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import org.apache.maven.shared.invoker.*;
+import org.junit.jupiter.api.Test;
+import java.io.File;
+import java.util.Arrays;
+
+public class MavenInvokerIntegrationTest {
```
jonesbusy, line 8: By default to be run by failsafe IT must match a certain pattern: https://maven.apache.org/surefire/maven-failsafe-plugin/examples/inclusion-exclusion.html#inclusions

plugin-modernizer-cli/src/test/java/io/jenkins/tools/pluginmodernizer/cli/MavenInvokerIntegrationTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Test
+    public void testMavenGoalExecution() throws MavenInvocationException {
+        InvocationRequest request = new DefaultInvocationRequest();
+        request.setPomFile(new File("pom.xml"));
+        request.setGoals(Arrays.asList("clean", "verify"));
```
jonesbusy, line 14: We should run not run clean verify on the root pom

But invoke something like `mvn exec:exec -Dexec.executable="java"`.....` by running the CLI previously build and packaged in a previous phase

Issue #420, opened 2024-12-08
Simplify usage of CLI


jonesbusy:
### What feature do you want to see added?

The CLI interface has too much parameter and we mix action and option.

I would suggest to have 1st grade command like

`plugin-modernizer list`
`plugin-modernizer dry-run`
`plugin-modernizer run`
`plugin-modernizer fetch-metadata`
`plugin-modernizer validate`

etc

This issue can be implemented in several PR

jonesbusy:
Partially implemented by https://github.com/jenkins-infra/plugin-modernizer-tool/pull/423

Need cleanup with option

PR #461, opened 2024-12-15
Split global options

jonesbusy:
Fix: https://github.com/jenkins-infra/plugin-modernizer-tool/issues/420

- Replace --clean-data by subcommand
- Removed some unused options

I also added a negative integration test that skip metadata when a plugin is deprecated

To have a positive test we should use https://testcontainers.com/modules/git/ to perform the clone of a fake plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,13 +94,6 @@ From there you need to save both ID of installation (found on URL)
 
 `https://github.com/organizations/<org>/settings/installations/<installation ID>`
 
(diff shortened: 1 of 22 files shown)


jonesbusy:
https://github.com/sparsick/testcontainers-git

Only side effect is that we will need to run CI on VM instead of container 

I'm pretty sure the git plugin is also using this module

Issue #460, opened 2024-12-15
Support for signed commits


jonesbusy:
### What feature do you want to see added?

SInce JGit version 7.1 it support SSH keys signing. 

Perhaps it could be the easiest solution rather than using GPG signing

To investigate

jonesbusy:
Will be probably one of my next task.

Some plugin require signed commits : https://github.com/jenkinsci/token-macro-plugin/pull/239

gounthar:
Would it be possible to include instructions in the documentation on how users can configure their SSH key for this tool?

Issue #318, opened 2024-10-10
Proposal: Enable Modernization of Plugins in Private Repositories


gounthar:
## Current Situation

Currently, our modernization process relies solely on the update center to retrieve plugin repositories for modernization efforts.

## Challenge

This approach limits our ability to modernize certain types of plugins:

1. Plugins with security issues that have been forked into private repositories for resolution
2. Plugins developed by private companies that are not (yet) part of the jenkinsci organization
3. Proprietary plugins that will remain outside the jenkinsci organization

## Proposed Solution

We propose expanding the modernization capabilities to include plugins in private repositories. This could be achieved by:

1. Allowing the use of repository URLs as input for the modernization process
2. Implementing a mechanism to authenticate and access private repositories using appropriate environment variables

## Implementation Considerations

To implement this feature, we would need to:

1. Modify the input process to accept repository URLs in addition to plugin names
2. Develop a secure method for handling authentication credentials (e.g., using environment variables)
3. Ensure the tool can clone, fork, and create pull requests for private repositories
4. Update documentation to reflect these new capabilities and explain the setup process

## Benefits

This enhancement would:

1. Increase the tool's flexibility and usefulness
2. Enable faster resolution of security issues in plugins
3. Support companies with proprietary plugins in keeping their extensions up-to-date

## Questions for Discussion

1. What security measures should we implement to protect authentication credentials?
2. How can we ensure the tool remains easy to use while adding this functionality?
3. Are there any potential drawbacks or risks associated with this approach?

We welcome your thoughts and feedback on this proposal to help shape the future development of our plugin modernization tool.

jonesbusy:
In theory it should be possible. Also since https://github.com/jenkins-infra/plugin-modernizer-tool/pull/458 we can add a custom API URL (for example if your org is using self hosted GitHub enterprise)

Different URL can be overriden by env vars or CLI args

JENKINS_UC
JENKINS_PLUGIN_INFO
JENKINS_PHS
JENKINS_PLUGINS_STATS_INSTALLATIONS_URL
GH_TARGET_ORGANISATION

I think this issue just deserve documentation and one validation 

I will most likely need to finish https://github.com/jenkins-infra/plugin-modernizer-tool/issues/420 to split global option and command option

jonesbusy:
Also now support local plugin via #503.

Integration test with testcontainer show that is possible to override all options yo.make it work with private plugin as long they are published on GitHub or GitHub enterprise

PR #463, opened 2024-12-15
Switch to SSH transport 

jonesbusy:
The goal for this PR is to allow support for SSH transport on JGit. In order to clone and perform Git operation via SSH

This require to pass the `--ssh-private-key`

This is also some preparation for https://github.com/jenkins-infra/plugin-modernizer-tool/issues/460

This also add better support for https://github.com/jenkins-infra/plugin-modernizer-tool/issues/318 because the authentication is passed also during fetch command

The PR also add and integration test for the `build-metadata` subcommand and a Git testcontainer

The PR also fixes various bug during testing.

One example is the support of `--maven-local-repo` which was not cutomizable. For example on ci.jenkins.io is different and was breaking the resolution of the core module (which was installed on different location `m2repo` of the workspace

The option is also useful in order to run the tool were the /home/ folder is not writable (typically on container or kubernetes distribution

One improvement is that we miss https://github.com/hub4j/github-api/pull/1996 in order to validate the target GH SSH keys

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,14 +3,13 @@ target/
 !**/src/main/**/target/
 !**/src/test/**/target/
 
-### Plugins Directory ###
-test-plugins
-
 ### Log file ###
 logs/
 
 ### IntelliJ IDEA ###
 .idea/
+!.idea/runConfigurations/
+!.idea/.gitignore/
 *.iws
 *.iml
 *.ipr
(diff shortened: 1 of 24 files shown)


jonesbusy:
Right now all keys are accepted https://github.com/apache/mina-sshd/blob/master/docs/client-setup.md#serverkeyverifier

We should validate againts known keys.

GitHub provide such keys via API, but it's missing on the java client

https://github.com/hub4j/github-api/pull/1996

gounthar:
Time for another PR to `hub4j/github-api` ? 😉 
I have not yet finished [mine](https://github.com/hub4j/github-api/pull/1994).

jonesbusy:
PR finally green

gounthar (approved):

PR #495, opened 2024-12-20
Copy metadata to cache instead of moving it

jonesbusy:
- Copy metadata to cache instead of moving it
- Add attribute on recipe to force the rebuild
- Store the metadata on the pom.xml so it can be consumed

Idea is to not "manually" perform metadata collection before a recipes but add is as a first recipe.

So that later recipes applied on the trees can lookup for the marker instead of relying of ready manually the data from any storage

The build-metadata` will remains to force the rebuild on populate the local storage

### Testing done

Automated tests

### Submitter checklist

### DIFF
--- a/plugin-modernizer-cli/pom.xml
+++ b/plugin-modernizer-cli/pom.xml
@@ -50,6 +50,10 @@
       <version>3.3.0</version>
     </dependency>
     <dependency>
+      <groupId>org.openrewrite</groupId>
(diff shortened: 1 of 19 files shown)


jonesbusy:
Still evaluating if the search flag is relevant to print the search result back to code.

I will most likely keep the Marker implementation for the plugin metadata and remove the search result

Issue #505, opened 2024-12-23
Windows distribution via scoop


jonesbusy:
### What feature do you want to see added?

The CLI is distributed by Homebrew which support Linux (and Windows users with WSL2) and MacOSX.

It would be great to not let Windows user left behind and allow to distribute the CLI via a windows package manager.

I woulds suggest scoop because it's very similar to Homebrew (a Tap is a Bucket, a Formala is an app manifest)

It also work with JRelease : https://jreleaser.org/guide/latest/reference/packagers/scoop.html

Good first issue if anyone is interested by packaging

CodexRaunak:
Hey @jonesbusy I am interested in working on this one

jonesbusy:
Sure feel free to open a PR

gounthar:
What about Chocolatey?

jonesbusy:
The point with Chocolatey is packages need to be packaged and pushed to https://push.chocolatey.org/ (A bit like maven central). So it means managing an API key

With scoop it's downloaded through anywhere (a bit like Homebrew) so we can reuse the uber jar already

What we miss is "just" creating the app manifest and updating the jrelease configuration

I think scoop is much simplier for now

gounthar:
This issue was discussed during [the Infra Weekly Meeting](https://community.jenkins.io/tag/sig-infra).

In Jenkins, we currently utilize package managers such as Chocolatey and Homebrew, alongside the default Windows package manager.

If we decide to incorporate Scoop, we, as a community, must commit to keeping it updated and functioning properly. I am concerned that our infra team may not have the necessary bandwidth to manage this additional responsibility effectively. Should we successfully maintain Scoop, it could then be feasible to consider migrating the official Jenkins clients to a Scoop registry.

jonesbusy:
Yes but here we are not talking about Jenkins for end user but a tool for plugin maintainer. I think it's a bit different.

Using scoop for one tool doesn't mean we should also use scoop to distribute Jenkins right ? 

For HomeBrew, Jenkins is not managed by the infra-team since it's packaged by a 3rd party (See https://www.jenkins.io/download/lts/macos/) right ? It's the same for scoop https://bjansen.github.io/scoop-apps/main/jenkins/. A 3rd party decided to package it and make it available (with a very outdated version by the way)

[...]

PR #515, opened 2024-12-27
Support for scoop distribution

CodexRaunak:
### Description
This PR adds support for distributing the plugin-modernizer plugin via Scoop on Windows. This integration leverages JReleaser to package and distribute the plugin, making it easy for Windows users to install and use the tool through Scoop.

Changes made:
- Added a Scoop bucket for distributing the plugin-modernizer plugin on Windows.
- Created an app manifest for the plugin and updated the JReleaser configuration to support Scoop as a distribution method.
- The link for the bucket where the app manifest is located is https://github.com/CodexRaunak/plugin-modernizer-bucket

Fix https://github.com/jenkins-infra/plugin-modernizer-tool/issues/505

### Testing done
This change was tested by successfully installing and running the plugin-modernizer on a Windows machine using Scoop. The installation process went through without errors.

To verify the installation:
I ran scoop install plugin-modernizer and confirmed that the installation completed successfully.

### Submitter checklist

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -99,9 +99,9 @@ jobs:
           JRELEASER_PROJECT_VERSION: ${{ steps.extract-version.outputs.version }}
           JRELEASER_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Generate Homebrew token
+      - name: Generate token
         uses: tibdex/github-app-token@v2
-        id: generate_homebrew_token
+        id: generate_token
         with:
           app_id: ${{ secrets.GORELEASER_APP_ID }}
           private_key: ${{ secrets.GORELEASER_APP_PRIVKEY }}
@@ -114,7 +114,8 @@ jobs:
         env:
           JRELEASER_PROJECT_VERSION: ${{ steps.extract-version.outputs.version }}
           JRELEASER_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          JRELEASER_HOMEBREW_GITHUB_TOKEN: ${{ steps.generate_homebrew_token.outputs.token }}
+          JRELEASER_HOMEBREW_GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
+          JRELEASER_SCOOP_GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
 
       - name: Archive output
         if: always()
--- a/jreleaser.yml
+++ b/jreleaser.yml
@@ -54,3 +54,18 @@ packagers:
     skipTemplates:
       - cask.rb.tpl
       - README.md.tpl
+
+  scoop:
+    active: RELEASE
+    repository:
+      active: RELEASE
+      owner: jenkins-infra
+      name: scoop-bucket  
+      tagName: '{{distributionName}}-{{projectVersion}}'
+    commitAuthor:
+      name: jenkins-infra
+      email: jenkins-infra-team@googlegroups.com
+    extraProperties:
+      skipLicenseFile: true
+    skipTemplates:
+      - scoop-bucket/README.md.tpl


pom.xml
```diff
@@ -39,7 +39,7 @@
     <maven.compiler.source>21</maven.compiler.source>
     <maven.compiler.target>21</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <changelist>999999-SNAPSHOT</changelist>
+    <changelist>1025.v6051110469dc</changelist>
```
jonesbusy, line 42: This is incorrect. The changelist is set via the `-Dset.changelist` when a release is built

jreleaser.yml
```diff
@@ -54,3 +54,18 @@ packagers:
     skipTemplates:
       - cask.rb.tpl
       - README.md.tpl
+
+  scoop:
+    active: RELEASE
+    repository:
+      active: RELEASE
+      owner: CodexRaunak
```
jonesbusy, line 62: Wrong owner

jonesbusy (changes requested):

jonesbusy:
Were you able to test the update on `https://github.com/CodexRaunak/plugin-modernizer-bucket/blob/main/plugin-modernizer.json` and the cd workflow ?

I dont' see any automatic commit there

.github/workflows/cd.yaml
```diff
@@ -113,6 +113,23 @@ jobs:
...
           JRELEASER_HOMEBREW_GITHUB_TOKEN: ${{ steps.generate_homebrew_token.outputs.token }}
+    
+      - name: Generate Scoop token
+        uses: tibdex/github-app-token@v2
+        id: generate_scoop_token
+        with:
+          app_id: ${{ secrets.SCOOP_APP_ID }}
+          private_key: ${{ secrets.SCOOP_APP_PRIVKEY }}
+
+      - name: Publish release to Scoop
```
jonesbusy, line 124: This is not how jreleaser is working. The publish will publish both to Homebrew and scoop. No need to call it twice 

Also you just need one token since we use the same service account/app ID for both. Only the repository is different

Why not just adding 

```
JRELEASER_SCOOP_GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
```

And rename the steps to somethink like `Generate token` instead of `Generate Homebrew token`

CodexRaunak: @jonesbusy should i rename the GORELEASER_APP_ID to JRELEASER_APP_ID or this is fine? Is this name used for consistency purpose across different tools?

jonesbusy: Sadly those one no. Because they are set at org level

Screenshot from 2024-12-30 16-35-42

jreleaser.yml
```diff
@@ -54,3 +54,18 @@ packagers:
     skipTemplates:
       - cask.rb.tpl
       - README.md.tpl
+
+  scoop:
+    active: RELEASE
+    repository:
+      active: RELEASE
+      owner: jenkins-infra
+      name: plugin-modernizer-bucket
```
jonesbusy, line 63: Can a bucket store more apps ? If yes perhaps we need to ask a repo with something more generic ?

Perhaps 'scoop-bucket' ? What do you think ?

CodexRaunak: Yeah, that would be better, we can store multiple app manifests into this single bucket, scoop-bucket will go fine.

jonesbusy (approved):

jonesbusy:
To merged when the repo is provisionned on jenkins-infra

I would also like to test on one bucket https://github.com/jonesbusy/scoop-bucket before

jonesbusy:
I've raised https://github.com/jenkins-infra/helpdesk/issues/4468

I've also tested the publish and result of https://github.com/jonesbusy/scoop-bucket/blob/main/bucket/plugin-modernizer.json

Sadly I don't have a Windows environment to test now but I think it's looking pretty neat

My only concern is this part

```json
  "checkver": {
    "url": "https://github.com/jonesbusy/plugin-modernizer-tool/releases/latest",
    "re": "v([\\d.]+).jar"
  },
```

I don't know if the updates will be detected ? Our versioning scheme like 1098.vd063a_8b_4fcdf will not match this regex

@CodexRaunak Did you test the update ? Do you know if we can customize this regex ?

It should be something like

`-([\\d.]+)\\.(.*).jar` (untested)

to detect the first part of the version which increment

CodexRaunak:
The above regex you mentioned `-([\\d.]+)\\.(.*).jar` needs a little correction on the `\\` part 
This `\\` is not detecting the version rather single \ is working `-([\d.]+)\.(.*).jar` fine.

https://regexr.com/ I tested the regex on this tester.
Also we can be bit more specfic than .*  like `([\d]+)\.([a-zA-Z0-9_]+)\.jar
`. These both are working fine but the lower one is more specefic.

Also it will select the `-` from the URL, so if we don't need the hyphen part just remove it from the regex start.

CodexRaunak:
@jonesbusy I have been testing the workflow, but facing an issue, do you have any idea why the CI test is not validating , and resolves to false
image
I believe its because of this line   `ID=$(gh api -X GET -F check_name="$NAME" -F status=completed /repos/$GITHUB_REPOSITORY/commits/$GITHUB_SHA/check-runs | jq -e '.check_runs | .[] | select(.conclusion == "success") | .id' || echo 'failed')` in the run.sh. I am stuck on trouble shooting it.

jonesbusy:
> The above regex you mentioned `-([\\d.]+)\\.(.*).jar` needs a little correction on the `\\` part This `\\` is not detecting the version rather single \ is working `-([\d.]+)\.(.*).jar` fine.
> 
> https://regexr.com/ I tested the regex on this tester. Also we can be bit more specfic than .* like `([\d]+)\.([a-zA-Z0-9_]+)\.jar `. These both are working fine but the lower one is more specefic.
> 
> Also it will select the `-` from the URL, so if we don't need the hyphen part just remove it from the regex start.

Thanks!

For the CI is because the workflow check the status on ci.jenkins.io. It will not work on your fork. You can only test some part of the workflow (like jreleaser)

CodexRaunak:
Ahh i see, well thanks

jonesbusy:
I confirm it work with current setup (Back to work, I can test on Windows)

scoop_update

When we have the bucket on jenkins-infra we can merge and test the CD workflow for and end2end test. But I'm quite confident it work

jonesbusy:
Merging: https://github.com/jenkins-infra/plugin-modernizer-tool/issues/505 was created.

Need to test end2end release

PR #541, opened 2024-12-30
Handle special case for analysis-pom-plugin parent

jonesbusy:
There is few plugin using a custom parent https://github.com/jenkinsci/analysis-pom-plugin/

### Testing done

Add tests

### Submitter checklist


jonesbusy:
Need probably more test. But at least now we are able to detect the correct parent and bom used by the analysis-pom

Issue #531, opened 2024-12-29
Recipe for Jenkinsfile updates 


jonesbusy:
### What feature do you want to see added?

Right now the recipes doesn't override the Jenkinsfile in order to not alternate exotic configuration.

We need more logic here when we modernize a plugin. For example 

- Add a java version with platform 
- remove totally a java version 
- add first level args, like useContainerAgent 

Etc.

See already groovy visitor that extract such information 

### Upstream changes

None

### Are you interested in contributing this feature?

None

jonesbusy:
https://github.com/jenkins-infra/pipeline-library/pull/897

As information related to implicit settings

basil:
> add first level args, like useContainerAgent

But only if the plugin is not running Docker-based tests. A few previous modernization PRs incorrectly added `useContainerAgent` and accidentally disabled Docker-based tests.

jonesbusy:
@basil Thanks, do you have any example of such recent PR ?

The current state is to not overwrite anything if a Jenkinsfile is present until we have logic : https://github.com/jenkins-infra/plugin-modernizer-tool/blob/a5c2da38530ccad409e22d7e6880514c141e3646/plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml#L26

I think we should be able to detect such types (pom, import) to know detect if we are using docker tests or not

The idea behind openrewrite is not not do any destructive changes (For example a Jenkinsfile adding configuration like https://github.com/jenkinsci/warnings-ng-plugin/blob/main/Jenkinsfile#L9-L11 should not lost it's information

basil:
> @basil Thanks, do you have any example of such recent PR ?

An example is https://github.com/jenkinsci/saml-plugin/pull/157, which (incorrectly) set `useContainerAgent: true`, thereby disabling Docker-based testing. I set `useContainerAgent: false` in https://github.com/jenkinsci/saml-plugin/pull/446 and confirmed that `LiveTest` was again running in CI builds. Fortunately, the test had not regressed during that time!

To be clear, it is always* OK to add an explicit `useContainerAgent: false` (*except if enforcing Spotless on Windows, pending https://github.com/jenkins-infra/helpdesk/issues/3865), but it is only OK to add an explicit `useContainerAgent: true` if there are no Docker-based tests.

jonesbusy:
Thanks. I almost forget about the spotless issue (that I was facing 1-2 times).

[...]

PR #544, opened 2024-12-31
Extract for info from Jenkinsfile and support different layout

jonesbusy:
Extract full platform information from Jenkinsfile and detect default values according to https://github.com/jenkins-infra/pipeline-library (including useContainerAgent or forkCount)

The only missing is the 'jenkinsVersion' that some plugin decide to override instead of using the default from the pom. I will add it on an other PR

This is also some preparation for https://github.com/jenkins-infra/plugin-modernizer-tool/issues/531 in order to keep the current config but migrate to new format. Or adding, removing JDK. Or removing extra jenkinsVersion or limit matrix size (for example 2 or 3 if necessary) to save compute resource

For example

```groovy
buildPlugin()
```

will have JDK8 on Linux and Windows

Or 

```groovy
buildPlugin(
   jdkVersions: [21, 17]
)
```

Will have Java 17 and 21 on Linux and Windows

This is some preparation to convert to the "new" format 

```groovy
buildPlugin(
  configurations: [
        [platform: 'linux', jdk: 21],
        [platform: 'windows', jdk: 17],
  ]
)
```

### Testing done

Mainly automated tests

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/JenkinsfileVisitor.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/JenkinsfileVisitor.java
@@ -1,7 +1,11 @@
 package io.jenkins.tools.pluginmodernizer.core.extractor;
 
 import io.jenkins.tools.pluginmodernizer.core.model.JDK;
+import io.jenkins.tools.pluginmodernizer.core.model.Platform;
+import io.jenkins.tools.pluginmodernizer.core.model.PlatformConfig;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Stream;
@@ -13,7 +17,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Visitor for a Jenkinsfile
+ * Visitor for a Jenkinsfile to accumulate @see PluginMetadata.
  */
 public class JenkinsfileVisitor extends GroovyIsoVisitor<PluginMetadata> {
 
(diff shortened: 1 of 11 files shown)


plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/JenkinsfileVisitor.java
```diff
@@ -36,67 +43,198 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations v
         return variableDeclarations;
     }
 
+    @Override
+    public J visitMapEntry(G.MapEntry mapEntry, PluginMetadata pluginMetadata) {
```
jonesbusy, line 47: I'm really not proud of this code.

We should use different visit callback and use cursor to get information higher on the tree. Not going down

But I least now I have test to validation behavior and will improve in the future

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/JenkinsfileVisitor.java
```diff
@@ -36,67 +43,198 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations v
...
-                    args.stream().flatMap(this::extractJdkVersions).distinct().toList();
+            // Empty args means Java 8 in Windows and Linux
+            if (args.size() == 1 && args.get(0) instanceof J.Empty) {
+                pluginMetadata.addPlatform(new PlatformConfig(Platform.LINUX, JDK.JAVA_8, null, true));
+                pluginMetadata.addPlatform(new PlatformConfig(Platform.WINDOWS, JDK.JAVA_8, null, true));
+                return method;
+            }
 
-            LOG.info("JDK versions found: {}", jdkVersions);
+            List<PlatformConfig> platforms = new LinkedList<>();
```
jonesbusy, line 92: I'm really not proud of this code.

We should use different visit callback and use cursor to get information higher on the tree. Not going down

But I least now I have test to validation behavior and will improve in the future

See `visitMapEntry` that only visit entries inside a method

```java
J.MethodInvocation method = getCursor().firstEnclosing(J.MethodInvocation.class);
``

PR #550, opened 2025-01-01
Remove relativePath precondition

jonesbusy:
In fact not needed. It think it was because initially plugin were fetch on a 'test-plugin' directory.

Additionally it add a new recipe to ensure the relative path on the parent

### Testing done

Automated tests and 

### Submitter checklist


gounthar:
I think it is needed, as in the `vagrant` plugin.
I'll give it another test.

jonesbusy:
Okay I will test with such plugin and restore it if needed. But I don't get why I work for some of them without the relative path. I don't know if there is an other condition that fail the download

PR #585, opened 2025-01-07
Add `MergeGitIgnoreRecipe` to merge existing .gitignore file with common entries

### LINKED ISSUE #532: Recipe for .gitignore update

Right now the .gitignore is just added when missing. 

We need some logic to merge with entries from archetype

### Upstream changes

None

_No response_

nagu165:
Added the logic to merge the existing gitignore files with the archetype gitignore files

Fixes  #532

### Testing done
Yes

-> Added the MergeGitIgnoreRecipe to handle the merging of the existing gitignore files with the archetype gitignore files.
-> Added tests to verify that the recipe is working as expected.
-> Logic such that the archetype file can be provided as a constructor argument and the recipe takes care of merging.

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MergeGitIgnoreRecipe.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MergeGitIgnoreRecipe.java
@@ -46,27 +46,35 @@ public class MergeGitIgnoreRecipe extends Recipe {
         @Override
         public PlainText visit(Tree tree, ExecutionContext ctx) {
             PlainText text = (PlainText) tree;
-            if (!(tree instanceof PlainText)) {
-                return null;
-            }
-            // Check if the current file is a `.gitignore` file
-            if (!text.getSourcePath().getFileName().toString().equals(".gitignore")) {
+            Path sourcePath = text.getSourcePath();
+
+            // Early return if source path is null
+            if (sourcePath == null) {
                 return text;
             }
 
-            Path gitIgnorePath = text.getSourcePath();
+            // Safely get filename with null checks
+            Path fileNamePath = sourcePath.getFileName();
+            if (fileNamePath == null) {
+                return text;
+            }
+
+            String fileName = fileNamePath.toString();
+            if (fileName == null || fileName.isEmpty() || !".gitignore".equals(fileName)) {
+                return text;
+            }
+
             try {
-                if (Files.exists(gitIgnorePath) && Files.exists(archetypeGitIgnorePath)) {
-                    String existingContent = Files.readString(gitIgnorePath);
+                if (Files.exists(sourcePath) && Files.exists(archetypeGitIgnorePath)) {
+                    String existingContent = Files.readString(sourcePath);
                     String archetypeContent = Files.readString(archetypeGitIgnorePath);
                     String mergedContent = mergeGitIgnoreFiles(existingContent, archetypeContent);
 
-                    // Write back the merged content to the file
-                    Files.writeString(gitIgnorePath, mergedContent);
-                    LOG.info("Merged .gitignore at {}", gitIgnorePath);
+                    Files.writeString(sourcePath, mergedContent);
+                    LOG.info("Merged .gitignore at {}", sourcePath);
                 }
             } catch (IOException e) {
-                LOG.error("Error processing .gitignore files at {}", gitIgnorePath, e);
+                LOG.error("Error processing .gitignore files at {}", sourcePath, e);
                 throw new RuntimeException("Failed to process .gitignore files", e);
             }
 
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MergeGitIgnoreRecipe.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MergeGitIgnoreRecipe.java
@@ -1,5 +1,10 @@
 package io.jenkins.tools.pluginmodernizer.core.recipes;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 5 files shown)


nagu165:
test was added only for the merging logic and not for the creation of gitignore file as it was mentioned that the creation of gitignore file already takes place if it didn't exist.so, I'm assuming it has already been taken care of

Thank you.

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MergeGitIgnoreRecipe.java
```diff
@@ -0,0 +1,92 @@
...
+    public String getDescription() {
+        return "Merges .gitignore entries from archetype with existing .gitignore file.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new GitIgnoreMerger(archetypeGitIgnorePath);
+    }
+
+    private static class GitIgnoreMerger extends MavenIsoVisitor<ExecutionContext> {
```
jonesbusy, line 39: Why a MavenIsoVisitor since we are manipulating raw text file ?

A TreeVisitor looks more appropriate

By using Files you bypass changing the LST during recipe execution.

You can check how other recipes are generating new files or modifying them.

plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/recipes/MergeGitIgnoreRecipeTest.java
```diff
@@ -0,0 +1,68 @@
...
+                *.iws
+                *.ipr
+                .idea
+                
+                # Eclipse project files
+                .settings
+                .classpath
+                .project
+                """,
+                sourceSpecs -> sourceSpecs.path("archetypes/common-files/.gitignore")
```
jonesbusy: This should not be part of the source files. I think the text file can be directly added as a constant on the recipe

plugin-modernizer-core/src/main/jte/pr-title-MergeGitIgnoreRecipe.jte
```diff
@@ -0,0 +1,5 @@
+@import io.jenkins.tools.pluginmodernizer.core.model.Plugin
```
jonesbusy, line 1: Missing test on TemplateUtils to ensure the PR title matches we expect

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MergeGitIgnoreRecipe.java
```diff
@@ -0,0 +1,89 @@
...
+            try {
+                if (Files.exists(gitIgnorePath) && Files.exists(archetypeGitIgnorePath)) {
+                    String merged = mergeGitIgnoreFiles(
+                            Files.readString(gitIgnorePath), Files.readString(archetypeGitIgnorePath));
+                    Files.writeString(gitIgnorePath, merged);
+                }
+            } catch (IOException e) {
+                LOG.error("Error processing .gitignore files", e);
+            }
+            return document;
```
jonesbusy, line 62: This literally ignore pretty much all errors. I don't think it's a good idea to not rethrow the exeption

jonesbusy (changes requested):
Need a second iteration. The visitor is wrong since it's implement a MavenIsoVisitor suited for pom.xml manipulation.

Perhaps missing some test for the recipe (.gitignore not exists, no changes done, entries in different order etc....)

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MergeGitIgnoreRecipe.java
```diff
@@ -46,27 +46,35 @@ private static class GitIgnoreMerger extends TreeVisitor<PlainText, ExecutionCon
         @Override
         public PlainText visit(Tree tree, ExecutionContext ctx) {
```
jonesbusy, line 47: In fact I found there is a more adapted visitor `PlainTextVisitor` more adapted than the TreeVisitor Did you found this as well ?

```java
PlainText visitText(PlainText text, ExecutionContext ctx) {
   ...
   return text.withText(mergedContent);
```

As I say OpenRewrite should not manipulate the files directly but it's LST. Then the LST is written back to source code

nagu165: > In fact I found there is a more adapted visitor `PlainTextVisitor` more adapted than the TreeVisitor Did you found this as well ?
> 
Yes, Sir I did find it but it wasn't used anywhere else in the project, so i was hesitant to use it.
> ```java
> PlainText visitText(PlainText text, ExecutionContext ctx) {
>    ...
>    return text.withText(mergedContent);
> ```
> 
> As I say OpenRewrite should not manipulate the files directly but it's LST. Then the LST is written back to source code

Understood sir.

jonesbusy:
Thanks for your effort. I think we are good to go.

We can still improve in the future if we discover some edge case

jonesbusy (approved):

nagu165:
> Thanks for your effort. I think we are good to go.
> 
I'm very grateful for your guidance sir, Thank you
> We can still improve in the future if we discover some edge case

Yes sir, I'd be happy to do so
Thank you.

jonesbusy:
Test shouldMergeGitIgnoreEntries is badly implemented. I miss it. It creates two .gitignore files instead of checking content before/after.

nagu165:
> Test shouldMergeGitIgnoreEntries is badly implemented. I miss it. It creates two .gitignore files instead of checking content before/after.

Yes sir, I get it now.
Thank you for pointing it out.

PR #607, opened 2025-01-10
Storing logs close to the sources 

### LINKED ISSUE #506: Allow debug logs into a writtable location

Right now log files are always created into the `logs` directory where the CLI is run

Is not convenient when running the CLI tool on a non-writtable location (like Kubernetes)

In addition when distributing the CLI (via Homebrew for example) it would be nice to write logs to a standard location (like /var/log)

Perhaps this could be configurable via an env var

### Upstream changes

_No response_

_No response_

CodexRaunak:
Fix #506 

Changes include:
Store the plugin logs to `/home/user/.cache/jenkins-plugin-modernizer-cli/<plugin>/logs/invoker.logs` and the default logs to `/home/user/.cache/jenkins-plugin-modernizer-cli/modernizer.logs`. In `getDiscriminatorValue()` build the plugin based on the name of the marker, and then we return the full absolute path for the plugin specefic log directory via the `getMarker().`

### Testing done
Updated the test cases.
Did the testing for `PluginLoggerDiscriminator `and of `getMarker()` method. 

### Submitter checklist

### DIFF
--- a/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/PluginLoggerDiscriminator.java
+++ b/plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/PluginLoggerDiscriminator.java
@@ -2,6 +2,10 @@ package io.jenkins.tools.pluginmodernizer.cli;
 
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.sift.AbstractDiscriminator;
+import io.jenkins.tools.pluginmodernizer.cli.options.GlobalOptions;
+import io.jenkins.tools.pluginmodernizer.core.config.Config;
+import io.jenkins.tools.pluginmodernizer.core.model.Plugin;
+import java.nio.file.Path;
 import java.util.List;
 import org.slf4j.Marker;
 
@@ -10,11 +14,23 @@ public class PluginLoggerDiscriminator extends AbstractDiscriminator<ILoggingEve
     @Override
     public String getDiscriminatingValue(ILoggingEvent iLoggingEvent) {
         List<Marker> markers = iLoggingEvent.getMarkerList();
+
+        Config.Builder builder = Config.builder();
+        GlobalOptions globalOptions = GlobalOptions.getInstance();
+        globalOptions.config(builder);
+        Config config = builder.build();
+
+        String cachePath = config.getCachePath().toString();
+
         if (markers == null || markers.isEmpty()) {
-            return "modernizer";
+            return Path.of(cachePath, "modernizer.logs").toString();
         }
+
         final Marker marker = markers.get(0);
-        return marker.getName();
+        String markerName = marker.getName();
+        Plugin plugin = Plugin.build(markerName);
+
+        return Path.of(cachePath, plugin.getLogFile().toString()).toString();
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


CodexRaunak:
can you please review it, so that I can work on it further

plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/PluginLoggerDiscriminator.java
```diff
@@ -11,10 +15,15 @@ public class PluginLoggerDiscriminator extends AbstractDiscriminator<ILoggingEve
     public String getDiscriminatingValue(ILoggingEvent iLoggingEvent) {
         List<Marker> markers = iLoggingEvent.getMarkerList();
         if (markers == null || markers.isEmpty()) {
-            return "modernizer";
+            return Path.of(System.getProperty("user.home"), ".cache", "jenkins-plugin-modernizer-cli", "modernizer.logs").toString();
```
jonesbusy, line 18: The problem is that is doesn't honor the Settings and Config for the --cache-path option.

Did some test where done with such option ?

I think we also need to update the integration test that create attachement for such logs

jonesbusy:
Can you run spotless un order to let the test run ?

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/model/Plugin.java
```diff
@@ -476,7 +478,22 @@ public Path getLogFile() {
      * @return Marker object
      */
     public Marker getMarker() {
-        return MarkerFactory.getMarker(name);
+        Path logsPath;
```
jonesbusy, line 481: Why we don't use getLogFile() from Plugin ? I would expect to see it updated and return the new location

jonesbusy (changes requested):

jonesbusy:
Thanks, it's a good start but need some polish to ensure we don't create regression and we ensure the --cache-path work as expected.

jonesbusy:
Looks good to me. Just ensuring test are passing. 

I just want to do some interactive test tomorrow and we should be good to merge.

The idea of the singleton for global is a good idea to get the value from CLI if its overriden.

jonesbusy:
Please ensure to fix spotless before pushing, or it will not run the test

jonesbusy:
Interesting the `testHelp` is failling. It seems it cause some regression

Additionally I run 

```
java -jar plugin-modernizer-cli/target/jenkins-plugin-modernizer-999999-SNAPSHOT.jar dry-run --cache-path my-cache --plugins pyenv-pipeline --recipe UpgradeToLatestJava8CoreVersion
```

And the modernizer.log and invoker.log are not moved to the my-cache but still on the ~/.cache/... etc...

CodexRaunak:
Now the logs are getting stored in the correct --cache path, I tested it, the commands were not using the singleton instance for the mixin before which was causing it. 

> Interesting the `testHelp` is failling. It seems it cause some regression
> 
> Additionally I run
> 
> ```
> java -jar plugin-modernizer-cli/target/jenkins-plugin-modernizer-999999-SNAPSHOT.jar dry-run --cache-path my-cache --plugins pyenv-pipeline --recipe UpgradeToLatestJava8CoreVersion
> ```
> 
> And the modernizer.log and invoker.log are not moved to the my-cache but still on the ~/.cache/... etc...

jonesbusy:
Lgtm, thanks for your contribution

jonesbusy (approved):

PR #630, opened 2025-01-13
Do not upgrade jenkins.version if not supported by Jenkinsfile

jonesbusy:
### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/recipes/UpgradeJenkinsVersionTest.java
+++ b/plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/recipes/UpgradeJenkinsVersionTest.java
@@ -1,8 +1,10 @@
 package io.jenkins.tools.pluginmodernizer.core.recipes;
 
+import static org.openrewrite.groovy.Assertions.groovy;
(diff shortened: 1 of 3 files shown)


plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/recipes/UpgradeJenkinsVersionTest.java
```diff
@@ -93,6 +95,59 @@ void testPerformUpgradeWithoutBom() {
                  """));
     }
 
+    @Test
+    void testNoUpgradeIfJenkinsFileDoesNotSupportJava17() {
```
jonesbusy, line 99: No changes is done there. Since JDK 8 is still referenced and not supported by 2.452.4

jonesbusy:
@gounthar This should probably not be merged but it would be the approach to not upgrade to a version not supported by the Jenkinsfile.

Sadly on OpenRewrite we cannot configure such thing as declarative precondition. Declarative precondition doesn't support scanning recipes.

Scanning recipes are mandatory to get information accross different LTS (groovy, maven etc...)

So if we want to do that we should add such precondition (can be maybe factorized) on other recipes.

So it's probably better to wait for https://github.com/jenkins-infra/plugin-modernizer-tool/pull/630

PR #633, opened 2025-01-14
Insert containerAgent and forkCount flags

jonesbusy:
Other PR for #531 that just implement the visitor. Then we will see that values we pass to the visitor depending on the recipe.

In think in any case we need to remove the `jenkinsVersions`, `jdkVersions` and `platforms` that will be replaced by the `configurations` block 

TODO for this PR still, add comments like https://github.com/jenkinsci/archetypes/blob/master/common-files/Jenkinsfile#L6-L7

### Testing done

Automated tests

### Submitter checklist


jonesbusy:
I think since the comment are added after the "," they must be added as a prefix of the next argument.

I'm trying to check if we can override the delimiter "," by ", // comment" when creating the LTS

I will manage this point on an other PR

PR #634, opened 2025-01-14
Insert comments and configuration on `buildPlugin`

jonesbusy:
I think now the visitor is ready to be used with recipes. There might be still some use case I forgot, but I think the formatting is good

I think it's worth using it https://github.com/jenkins-infra/plugin-modernizer-tool/pull/630 to update the Jenkinsfile when we bump the jenkins.version (instead of doing nothing) So that we know what are the supported jenkins versions.

If a Jenkins version support 3 Java version perhaps we can only use the 2 latest one (For example on 2.454.4, Java 11, Java 17 and Java 21 are supported but we only build on 17 and 21).

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/model/PlatformConfig.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/model/PlatformConfig.java
@@ -1,7 +1,42 @@
 package io.jenkins.tools.pluginmodernizer.core.model;
 
+import java.util.List;
+
 /**
  * Extracted platform from Jenkinsfile
  */
-public record PlatformConfig(Platform name, JDK jdk, String jenkins, boolean implicit) {}
+public record PlatformConfig(Platform name, JDK jdk, String jenkins, boolean implicit) {
+
+    /**
+     * Get the default platform configurations
+     * @return the default platform configurations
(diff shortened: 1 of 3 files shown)


gounthar:
Sounds good to me.

Shouldn't we use the newest version that the plugin successfully compiled with? For example, if the plugin depends on Jenkins `2.454.4` but is compiled with Java 21, shouldn't we keep 17 and 21 and drop 11?

jonesbusy:
> UpgradeToLatestJava11CoreVersion

Considering that in 1 month when the 2.492.1 will be release the UpgradeToLatestJava11CoreVersion will also be the recommanded version to build.

But yes that's something to think about. We now have all the data needed to perform pretty much what we want depending on the jenkins version we target.

There must be a bit more logic here : https://github.com/jenkins-infra/plugin-modernizer-tool/pull/630/files#diff-f508ef08a0085fe4f53d36db9ac009b5ca5fa83cf293ef3dfdcfcdbcea65a3b4R65-R74

So select based on the `List<JDK> jenkinsJdks = JDK.get(minimumVersion)` which version we choose to insert on the Jenkinsfile

Also we probably need to see which platform we use depending on what the maintainer as choosen (maybe only build on Windows or Linux). So probably here keep the existing platform and just increment the JDK if needed and keep only the 2 most recent ?

Issue #537, opened 2024-12-30
Implement GitHub action


jonesbusy:
### What feature do you want to see added?

- action can be added on same repo and uses same version scheme 
- java based temurin 17
- download jar via java 
- dry-run option as input 
- recipe as input 
- GH_TOKEN as auth (API and git)
- against current repo (--plugin-path option ? Or do we need and option to run on current repo and not create fork ?
- test against one plugin

Probably very similar to 

https://github.com/jreleaser/release-action

### Upstream changes

See if we need an option to apply changes on current repo and not perform forking

morishima-senpai:
Hi, I would like to work on this issue:)

jonesbusy:
Many outdated plugin are not supporting yet Java 17 or 21. But I think the use case here is more about applying latest archetype changes on plugins. So I would expect the action/workflow to be installed on only active plugin.

The minimum Java version for Jenkins is 11 and 17 for recent LTS version.

So I would for for Java 17 to run plugin-modernizer action

gounthar:
@nagu165 : I think the question is not whether plugins will support Java 21 or not, but let me clarify a few things first. Many plugins won’t support Temurin 21 yet; we’re not ready. Jenkins core has been supporting Java 21 for a while now, and we started a campaign last year to upgrade the most installed plugins to Java 21. However, there are still many plugins that aren’t ready for Java 21, which is why we're focusing on getting them to work with Java 17 first.

Our goal in addressing this issue is to create a GitHub Action that would launch the modernized plugins. Then, the infra team could add this GitHub Action to existing plugins by default, or we could propose it through a PR to encourage maintainers to add this GitHub Action.

nagu165:
> Many outdated plugin are not supporting yet Java 17 or 21. But I think the use case here is more about applying latest archetype changes on plugins. So I would expect the action/workflow to be installed on only active plugin.
> 
> The minimum Java version for Jenkins is 11 and 17 for recent LTS version.
> 
> So I would for for Java 17 to run plugin-modernizer action

Ok sir, Understood.

nagu165:
[...]

PR #686, opened 2025-01-22
Add Github action using plugin-moderniser-tool

yashpal2104:
### Testing done

Add Github action using plugin-moderniser-tool

### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

PR #295, opened 2024-10-05
Support GitHub app authentication

jonesbusy:
There are still few things to take care (like granular permissions or git push) but now the CLI support GitHub app

### Testing done

Screenshot from 2024-10-05 18-56-43

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #687, opened 2025-01-22
Implemented GitHub actions

shinigami-777:
Towards https://github.com/jenkins-infra/plugin-modernizer-tool/issues/537.

Implemented the github action for plugin modernizer tool.
The `.github/workflow/test-modernizer-action.yml` file contains the test for validating it.
The input parameters are `command` (required) `arguments` (optional) and `working-directory`(optional).

### Submitter checklist

### DIFF
--- a/.github/actions/modernizer-action/action.yml
+++ b/.github/actions/modernizer-action/action.yml
@@ -16,6 +16,10 @@ inputs:
     required: false
     default: ${{ github.workspace }}
 
+  version:
+    description: 'The version of Plugin Modernizer to download and execute'
+    default: "latest"
+    required: true
 
 runs:
   using: "composite"
@@ -32,7 +36,11 @@ runs:
       shell: bash
       working-directory: ${{ inputs.working-directory }}
       run: |
-        API_URL="https://api.github.com/repos/jenkins-infra/plugin-modernizer-tool/releases/latest"
+        if [ "$VERSION" == "latest" ]; then
+          API_URL="https://api.github.com/repos/jenkins-infra/plugin-modernizer-tool/releases/latest"
+        else
+          API_URL="https://api.github.com/repos/jenkins-infra/plugin-modernizer-tool/releases/tags/${{ inputs.version }}"
+        fi
         JAR_URL=$(curl -s $API_URL | grep "browser_download_url" | grep ".jar" | head -n 1 | cut -d '"' -f 4)
         curl -L -o plugin-modernizer.jar $JAR_URL
 
--- a/.github/actions/modernizer-action/action.yml
+++ b/.github/actions/modernizer-action/action.yml
@@ -1,0 +1,40 @@
+name: "Plugin Modernizer Action"
+description: "Run the Plugin Modernizer tool on the Jenkins plugin"
+inputs:
+  command:
+    description: "Command to execute for Plugin Modernizer"
+    required: true
+    default: "validate"
+
+  arguments:
+    description: "The command-line arguments to be passed"
+    required: false
+    default: ""
+
+  working-directory:
+    description: "The directory to change into"
+    required: false
+    default: ${{ github.workspace }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Temurin 17
+      uses: actions/setup-java@v4
+      with:
+        java-version: 17
+        distribution: temurin
+
+    - name: Download Plugin Modernizer JAR
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: |
+        API_URL="https://api.github.com/repos/jenkins-infra/plugin-modernizer-tool/releases/latest"
+        JAR_URL=$(curl -s $API_URL | grep "browser_download_url" | grep ".jar" | head -n 1 | cut -d '"' -f 4)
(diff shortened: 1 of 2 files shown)


shinigami-777:
Thanks a lot for the references @jonesbusy  @gounthar .

Currently the test  from  `.github/workflow/test-modernizer-action.yml` is failing due to not setting of `maven home`. Should I change it to install maven for setting of the `maven home` variable? or is there a more suitable way with the temurin setup.

maven

jonesbusy:
Thanks! Looks good so far! For maven let's reuse the setup maven action I'm the test workflow. 

We can still tweak it after all,  but a test that run the validate command is a very good start

.github/actions/modernizer-action/action.yml
```diff
@@ -0,0 +1,40 @@
...
+
+  working-directory:
+    description: "The directory to change into"
+    required: false
+    default: ${{ github.workspace }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Temurin 17
```
jonesbusy, line 22: This can be replaced by

```yaml
- name: Setup Maven Action
  uses: s4u/setup-maven-action@v1.18.0
  with:
    java-distribution: 'temurin'
    java-version: 17
    maven-version: 3.9.9
    cache-enabled: true
```

jonesbusy (changes requested):

.github/actions/modernizer-action/action.yml
```diff
@@ -0,0 +1,40 @@
...
+      uses: actions/setup-java@v4
+      with:
+        java-version: 17
+        distribution: temurin
+
+    - name: Download Plugin Modernizer JAR
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: |
+        API_URL="https://api.github.com/repos/jenkins-infra/plugin-modernizer-tool/releases/latest"
```
jonesbusy, line 32: I'm wondering if we always need to use latest

Perhaps it's worth adding a parameter with optional version

```yaml
  version:
    description: 'The version of Plugin Modernizer to download and execute.'
    default: 'latest'
    required: true
```

What do you think?

jonesbusy (changes requested):

.github/actions/modernizer-action/action.yml
shinigami-777: Yeah it would be great to have an optional version parameter . Working on implementing it.

shinigami-777:
I have added the `version` parameter (set to latest by default).
Also made a few changes like setting the `maven home` and repository (necessary to run the jar) after testing a bit using a different branch. The  `.github/workflow/test-modernizer-action.yml` test runs without any problems now.

jonesbusy:
Thanks! LGTM

yashpal2104:
> Seems the test workflow fail on main. It's possible that GITHUB_TOKEN need more permission to get user information (if possible)

Maybe if the user does not have GitHub token we can pull the latest docker image for use of plugin-moderniser-tool
I have referenced it in my draft PR #686

jonesbusy:
How it will solve the issue? This is a temporary OIDC token with limited permission.

Running the job on docker will not change auth permission

jonesbusy:
Other option is #295 which maybe need rework since we need perhaps less complexity when opening PR on the current repository and don't need forking accross organisations 

If we cannot fix the Github token permission I would suggest to change temporarily the command from validate to version to avoid the workflow to fail

shinigami-777:
> Seems the test workflow fail on main. It's possible that GITHUB_TOKEN need more permission to get user information (if possible)

@jonesbusy Can you check once if the `GH_TOKEN` variable is set in the secrets of this repository. I had used `secrets.GH_TOKEN` in the code since i made it with the one i had forked. (might have been better to use `GITHUB_TOKEN` didn't think much of it at that time).

If `GH_TOKEN` is not set then that might be the cause of failure. We can try replacing [this](https://github.com/jenkins-infra/plugin-modernizer-tool/blob/main/.github/workflows/test-modernizer-action.yml?#L18) line with `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} `

jonesbusy:
@shinigami-777 I see. Now I understand why it was working on your fork. This is indeed not possible. We should not use a GITHUB_TOKEN linked to a human user but linked to an app installation 

Each job generate a github token (valid during job execution) with limited permission. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token to change permission

If plugin modernizer need to access some API not granted by any of the scope, perhaps we need to adapt the failing part (I suspect we try to determine the email of the user, which might not work in an action)

The last solution if we cannot make it work with a GITHUB_TOKEN is https://github.com/jenkins-infra/plugin-modernizer-tool/pull/295. It's not battle tested and I suspect also some adjuestement. This last option would require to 

- Create a app
- Each maintainer need to install the app on their repository

So it's more effort of maintenance.

I will open an other issue to track such permission 

In the meanwhile can we change the default command to 'version' instead of 'validate?

Thanks

jonesbusy:
Created https://github.com/jenkins-infra/plugin-modernizer-tool/issues/692

PR #702, opened 2025-01-24
Implemented a small version for issue#688

nagu165:
updated updatecli.d/jenkins-lts.yaml to update the jenkins recommended version
### Testing done

Added 2 more targets in the  updatecli.d/jenkins-lts.yaml to update the jenkins versions in recipes.yml
With the further input from the maintainers would like to work on it further.

### Submitter checklist
- [x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x ] Ensure that the pull request title represents the desired changelog entry
- [x ] Please describe what you did


nagu165:
Hello sir, Just wanted your input regarding these changes, would like to work on this further.
Please do tell me if i'm missing something.
Thank you.

jonesbusy:
What if we use a comment to match the version? "#recommanded version managed by updatecli"? 

But I think a maven property with filtering on the resource looks more clean

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/config/RecipesConsts.java
```diff
@@ -19,4 +19,5 @@ private RecipesConsts() {
     public static final String OLD_BOM_VERSION_PATTERN =
             ""; // Old release line like 2.150.x, no need to filter by pattern
     public static final String INCREMENTAL_REPO_ID = "incrementals";
+    public static final String JENKINS_CORE_VERSION = "${jenkins.core.minimum.version}";
```
jonesbusy, line 22: Why not `Settings.getJenkinsMinimumVersion()` instead of filtering java source?

nagu165:
Hello sir, As per your suggestion i'm trying to use maven filtering on the recipes.yml to dynamically update the version but filtering isn't working as expected, Please share any info on why this is so.
Thank you

plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
```diff
@@ -148,7 +148,7 @@ recipeList:
       artifactId: plugin
       newVersion: 5.X
   - io.jenkins.tools.pluginmodernizer.core.recipes.UpgradeJenkinsVersion:
-      minimumVersion: 2.479.1
+      minimumVersion: "${jenkins.core.minimum.version}"
```
jonesbusy, line 151: This is not the minimum version. This value should not be changed since it's the first LTS supporting Java 17 and required for parent 5.x upgrade.

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/config/RecipesConsts.java
nagu165: Sorry sir, I missed it.
updating now.
Thank you.

plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
jonesbusy: Only value https://github.com/jenkins-infra/plugin-modernizer-tool/pull/685/files must be filtered/changed

plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
```diff
@@ -442,7 +442,7 @@ recipeList:
       artifactId: plugin
       newVersion: 4.51 # See https://www.jenkins.io/blog/2022/12/14/require-java-11/
   - io.jenkins.tools.pluginmodernizer.core.recipes.UpgradeJenkinsVersion:
-      minimumVersion: 2.346.3
```
jonesbusy, line 445: This is NOT the minimum version. It's the last LTS supporting Java 8. This value will never changes

plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
nagu165: In the example that sir gave i.e #685 this was also modified so i was testing with this.
Now i understand the issue.
Thanks.

jonesbusy: Only 2.452.4 was incremented to 2.462.3. The rest of value must not be changed

plugin-modernizer-core/pom.xml
```diff
@@ -218,12 +218,19 @@
...
       <resource>
         <targetPath>META-INF/rewrite</targetPath>
-        <filtering>false</filtering>
+        <filtering>true</filtering>
         <directory>src/main/resources/META-INF/rewrite</directory>
         <includes>
           <include>recipes.yml</include>
         </includes>
       </resource>
+      <resource>
```
jonesbusy, line 227: To be removed

plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
nagu165: Yes sir, I misunderstood previously
Now i understand my mistake
Thank you.

jonesbusy:
> Hello sir, As per your suggestion i'm trying to use maven filtering on the recipes.yml to dynamically update the version but filtering isn't working as expected, Please share any info on why this is so.
> Thank you

I do not know if you don't share details

plugin-modernizer-core/src/main/resources/META-INF/rewrite/recipes.yml
```diff
@@ -416,7 +416,7 @@ recipeList:
   - io.jenkins.tools.pluginmodernizer.UpgradeParentVersion
   - io.jenkins.tools.pluginmodernizer.AddPluginsBom
   - io.jenkins.tools.pluginmodernizer.core.recipes.UpgradeJenkinsVersion:
-      minimumVersion: 2.462.3
+      minimumVersion: "${jenkins.core.minimum.version}"
```
jonesbusy, line 419: I don't see it in the pom

nagu165:
Hello sir, in versions.properties - openrewrite.maven.plugin.version = ${openrewrite.maven.plugin.version} is causing circular refernce error so i replaced it with a constant for now, How to modify it properly without any breaking changes.
Thank you.

nagu165:
Is this change okay?

jonesbusy:
Tests are failing

plugin-modernizer-core/src/main/resources/versions.properties
```diff
@@ -1,4 +1,4 @@
-openrewrite.maven.plugin.version = ${openrewrite.maven.plugin.version}
+openrewrite.maven.plugin.version = 5.10.0
```
jonesbusy, line 1: This is not correct. We need to continue filtering from the pom

plugin-modernizer-core/pom.xml
```diff
@@ -227,6 +227,26 @@
     </resources>
     <plugins>
 
+      <plugin>
```
jonesbusy, line 230: What is the reason to override the properties-maven-plugin rather than adding a new <resource> ?

nagu165: But i did add a unique execution id so, won't it run concurrently instead of overriding?

nagu165:
Hello sir, May i please know why the tests might be failing on linux, I'm stuck trying to troubleshoot based on the error logs.
Thank you.

nagu165:
Hello sir, The tests are failing on linux
Maybe because of the change intoduced in the pom, do i need to modify the test cases in CommandLineITCase.java or is there another approach as i'm stuck troubleshooting this and couldn't find a better solution yet.
Thank you.

jonesbusy:
Superseded by https://github.com/jenkins-infra/plugin-modernizer-tool/pull/776

Issue #104, opened 2024-07-17
Optimize Repository Forking Process


gounthar:
## Current Situation
When forking a repository on GitHub, our current process retrieves all branches from the original repository.

## Issue
This approach may be inefficient and unnecessary for our use case.

## Desired Outcome
We aim to fork only the default branch of the repository, as this is typically sufficient for our work.

## GitHub GUI Option
The GitHub web interface provides an option to fork only the default branch (usually the main branch).

## JGit Consideration
It's unclear whether JGit, the Java library we use for Git operations, offers a similar option to fork only the default branch.

## Proposed Solution
1. Investigate if JGit provides functionality to fork only the default branch.
2. If available, implement this feature in our forking process.
3. If not available through JGit, explore alternative methods to achieve this goal.

## Benefits
1. **Efficiency**: Reduces unnecessary data transfer and storage.
2. **Simplicity**: Maintains a cleaner repository structure in our forks.
3. **Focus**: Aligns with our primary need of working with the main branch.

## Next Steps
1. Research JGit documentation and community resources for potential solutions.
2. If no direct JGit solution exists, consider implementing a custom approach to achieve the desired outcome.
3. Test and validate any new implementation to ensure it doesn't introduce unforeseen issues.

## Conclusion
Optimizing our forking process to retrieve only the main branch will streamline our workflow and improve overall efficiency. This change aligns with best practices for managing forks when only the main branch is needed for development or analysis purposes.

lemeurherve:
You might not even need all commits from the primary branch (which can be
"master") but only the last one.

gounthar:
Yes indeed.

SohamJuneja:
@gounthar I explored about JGit , their docs and community forum and this can be done by JGit , it  offers an option like this:
CloneCommand.setCloneAllBranches(false)

gounthar:
Thanks @SohamJuneja. 👍 
I think @jonesbusy implemented it since then. 🤔

SohamJuneja:
@gounthar ohh , didnt check this🥹 , why's the issue still open then? Is there anything needed to be done here?

gounthar:
My bad, I just launched the tool on pipeline-graph-view, and it forked with the 12 branches. 🤷 
https://github.com/gounthar/pipeline-graph-view-plugin

SohamJuneja:
I guess there's a simple fix , just need to add one line in GHServiceTest.java 
[...]

PR #710, opened 2025-01-25
Upgrade to GitHub API 2.x

jonesbusy:
Update GitHub API to 2.0-alpha-3

### Testing done

Automated tests

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #730, opened 2025-01-27
Optimize by forking Default Branch Only

shinigami-777:
Towards #104 
Set the `defaultforkonly` to true with `GHRepositoryForkBuilder` to in place of `fork()` (fork was also deprecated in the recent release) to optimize forking process.

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/github/GHService.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/github/GHService.java
@@ -360,10 +360,10 @@ public class GHService {
             LOG.info(
                     "Forking the repository to personal account {}...",
                     getCurrentUser().getLogin());
-            return defaultbranchonlyfork(originalRepo, null);
+            return fork(originalRepo, null);
         } else {
             LOG.info("Forking the repository to organisation {}...", organization.getLogin());
-            return defaultbranchonlyfork(originalRepo, organization);
+            return fork(originalRepo, organization);
         }
     }
 
@@ -376,7 +376,7 @@ public class GHService {
      * @throws IOException          If the fork operation failed
      * @throws InterruptedException If the fork operation was interrupted
      */
-    private GHRepository defaultbranchonlyfork(GHRepository originalRepo, GHOrganization organization)
+    private GHRepository fork(GHRepository originalRepo, GHOrganization organization)
             throws IOException, InterruptedException {
         GHRepositoryForkBuilder builder = originalRepo.createFork();
         if (organization != null) {


shinigami-777:
The new version [github-api-2.0-alpha-3](https://github.com/hub4j/github-api/releases/tag/github-api-2.0-alpha-3) is released now. I took a try at solving the issue. But the `originalRepo.createFork() `  returns null giving the error
`Cannot invoke "org.kohsuke.github.GHRepositoryForkBuilder.organization(org.kohsuke.github.GHOrganization)" because "builder" is null`
I can't seem to understand why this is happening. Please take a look.

jonesbusy:
> Please take a look

Adapt the unit tests maybe?

shinigami-777:
> Adapt the unit tests maybe?

Sorry, I somehow completely overlooked this before.
Adjusted them for GHServiceTest.

shinigami-777:
@jonesbusy When I clone the main branch fresh, the tests are failing the same as [this](https://github.com/jenkins-infra/plugin-modernizer-tool/pull/730/checks?check_run_id=36243401412).

jonesbusy:
See #733

shinigami-777:
Thanks. I'm looking into it

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/github/GHService.java
```diff
@@ -359,13 +360,32 @@ private GHRepository forkRepository(GHRepository originalRepo, GHOrganization or
...
+    /**
+     * Fork the default branch only
+     *
+     * @param originalRepo The original repository to fork
+     * @param organization The organization to fork the repository to. Can be null for personal account
+     * @return The forked repository
+     * @throws IOException          If the fork operation failed
+     * @throws InterruptedException If the fork operation was interrupted
+     */
+    private GHRepository defaultbranchonlyfork(GHRepository originalRepo, GHOrganization organization)
```
jonesbusy, line 379: Normally we just snake case for method name `defaultBranchOnlyFork`.

Why not just `fork` since it's a private method?

jonesbusy:
Hi, I think the method name `defaultbranchonlyfork` should be renamed to something like `fork`. That's enough since it's a private method and there will not be other way to fork

Then I think we are good to merge

Supersedes https://github.com/jenkins-infra/plugin-modernizer-tool/pull/710

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/github/GHService.java
shinigami-777: Yes fork would be good

jonesbusy (approved):

Issue #548, opened 2025-01-01
Scanning recipe to add dependency on detached plugin if package is used


jonesbusy:
### What feature do you want to see added?

When running `UpgradeToLatestJava8CoreVersion`

Several issue were found on very oudated plugins see https://github.com/jenkins-infra/plugin-modernizer-tool/pull/547

This task is to implement a scanning recipe to add a detached plugin 

See : https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/jenkins/split-plugins.txt

I guess the list could be stored in an enum. I don't think we need to dynamically fetch it and parse it. I don't expect much more plugin to be detached in the future

In the mentionned case command-launcher 2.86 1.0 was detached on 2.86 causing the modernization to fail due to missing dependency

CodexRaunak:
@jonesbusy can you please remove this line 111 `document.printAll()` from`AddDetachedPluginDependency`recipe. It prints the whole document, just for debugging purpose, actually I was left do some touch ups later but the PR got merged.

jonesbusy:
Feel free to submit an other PR if there is more cleanup to perform

PR #759, opened 2025-02-01
Scanning recipe to add dependency on detached plugin

CodexRaunak:
Issue #548
Stored all the details of the split plugins in an `enum`, with the name of their `classes`.
Scanning recipe matches any of the plugins used from the `DetachedPlugins`, if found then it checks if the `jenkins version > last core release` of that plugin which means now it has been detached, so it adds it in the `accumulator`.
Then in the visitor phase we `add the dependency` of the detached plugin.

### Testing done
Added the unit test for the recipe and tested it with `UpgradeToLatestJava8CoreVersion`.

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/AddDetachedPluginDependency.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/AddDetachedPluginDependency.java
@@ -1,0 +1,116 @@
+package io.jenkins.tools.pluginmodernizer.core.recipes;
+
+import io.jenkins.tools.pluginmodernizer.core.model.DetachedPlugins;
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.maven.artifact.versioning.ComparableVersion;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.ScanningRecipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.maven.AddDependency;
+import org.openrewrite.maven.MavenIsoVisitor;
+import org.openrewrite.xml.tree.Xml;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AddDetachedPluginDependency extends ScanningRecipe<Set<String>> {
+
+    /**
+     * Logger
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(AddDetachedPluginDependency.class);
+
+    /**
+     * The jenkins version.
+     */
+    @Option(displayName = "Version", description = "Jenkins version.", example = "2.440.3")
+    String jenkinsVersion;
+
+    /**
+     * Constructor.
+     * @param jenkinsVersion The Jenkins version.
+     */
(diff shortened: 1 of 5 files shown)


CodexRaunak:
Need a review on this PR, I am concerned about the enum part, is storing the class names of the plugins to match the plugin used or not, the correct way?

CodexRaunak:
image

Also about this note, all the plugins between and including `matrix-auth` and `jaxb` have the incorrect `last core release` as mentioned.
So i need to change the enum `one version prior of the last core release.` As last core release defines the plugin is supported till that release.

jonesbusy:
> Need a review on this PR, I am concerned about the enum part, is storing the class names of the plugins to match the plugin used or not, the correct way?

Thanks for your PR. I would say if we can the package name. If we have some conflict (like LDAP) We could use class name.

I will need some time to review it but looks a very good start.

Can we have a test for every plugin?

jonesbusy:
> image
> 
> Also about this note, all the plugins between and including `matrix-auth` and `jaxb` have the incorrect `last core release` as mentioned.
> So i need to change the enum `one version prior of the last core release.` As last core release defines the plugin is supported till that release.
> 

I don't have more information about that. But I guess it's no so important since se will always modernizer to a much more recent versions

CodexRaunak:
> I don't have more information about that. But I guess it's no so important since se will always modernizer to a much more recent versions

Currently what I am doing is, I am checking if a plugin let's say `InstanceIdentity` is being used in a project. Then I am not directly adding it to the dependency, because maybe the jenkins version we are upgrading to may not have detached it. So that's why the last core release is maybe important to us, beacause it will help us check that we actually want to add it as a dependency or not.

If a plugin say `InstanceIdentity` is being used and its `lastCoreRelease` is `2.356`, so we only need to add it in dependency if the `jenkins version` is > `2.356`. If the `jenkinsVersion `is less than equal to `2.356 `don't add it in dependency as its not detached for that jenkins.

CodexRaunak:
I added the some test cases for every plugin, 
In the enum we are storing the `packageName `also, but if it has a conflict i.e more than one plugin use the same package then I added the core class names of that plugin.
Most of the plugins are overlapping so there package info is empty.

plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/AddDetachedPluginDependency.java
```diff
@@ -0,0 +1,116 @@
...
+        return new MavenIsoVisitor<>() {
+            @Override
+            public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
+                for (String pluginId : acc) {
+                    DetachedPlugins plugin =
+                            DetachedPlugins.valueOf(pluginId.toUpperCase().replace("-", "_"));
+                    document = (Xml.Document) new AddDependency(
+                                    plugin.getGroupId(),
+                                    plugin.getPluginId(),
+                                    "RELEASE",
```
jonesbusy, line 98: Is the "latest.release" working?

CodexRaunak: Works only on some of them , not all

jonesbusy:
Thanks! LGTM. We can still improve in the future if we discover some border case

jonesbusy (approved):

Issue #771, opened 2025-02-05
.cache/jenkins-plugin-modernizer-cli/log-parser/sources/src/main/java/hudson/plugins/logparser/LogParserAction.java:[135,33] incompatible types: org.kohsuke.stapler.StaplerRequest2 cannot be converted to org.kohsuke.stapler.StaplerRequest


CodexRaunak:
### Plugin Modernizer version

plugin modernizer 999999-SNAPSHOT (2025-02-05T12:58:22Z)

### What Operating System are you using

Windows-10

### Reproduction steps

```
java -jar plugin-modernizer-cli/target/jenkins-plugin-modernizer-999999-SNAPSHOT.jar dry-run --plugins log-parser --recipe UpgradeNextMajorParentVersion
Run Plugin Modernizer in dry-run mode 
Fetching plugin code locally log-parser... 
locking FileBasedConfig[C:\Users\DELL PC\.config\jgit\config] failed after 5 retries 
Fetched repository from https://github.com/jenkinsci/log-parser-plugin.git to branch refs/heads/develop 
Collecting metadata for plugin log-parser... Please be patient 
Done 
Running recipes io.jenkins.tools.pluginmodernizer.UpgradeNextMajorParentVersion for plugin log-parser... Please be patient 
Done 
Using minimum JDK 11 from metadata for plugin log-parser
Build failed 
Skipping formatting for plugin log-parser as it is not using Spotless
Verifying plugin log-parser with JDK 11... Please be patient
Build failed 
Done
Plugin log-parser failed to verify with JDK 11
Plugin log-parser verified successfully with JDK 11
Collecting metadata for plugin log-parser... Please be patient 
Build failed 
Done
Failed to collect metadata for plugin log-parser. Not retrying.
*************
Plugin: log-parser
Error: Build failed
*************

```

In the logs the build is failing due to 
`COMPILATION ERROR :  
2025-02-05T12:07:19.161Z [INFO] [Thread=StreamPumper-systemOut] - i.j.t.p.core.impl.MavenInvoker # [INFO] ------------------------------------------------------------- 
[...]

PR #774, opened 2025-02-05
Scanning recipe to check ChartUtil is used or not when migrating to Stapler2 and Javax to Jakarta

CodexRaunak:
#771 

Added a scanning recipe which scans the code base for usage of `ChartUtil`, if used it adds it in the accumulator.
In the editing phase we only perform the migration when `ChartUtil `is not used (i.e when accumulator is empty).

I confirmed that the `UpgradeNextMajorParentVersion` recipe is working on `log-parser`. 
As we are not  upgrading the `Stapler` in log-parser due to `ChartUtil` we are also not upgrading `javax` to `jakarta` as the `StaplerResponse `internally uses `javax.servelet`. So either if we perform the migrations as a whole or we don't.

```
Done
Plugin log-parser failed to verify with JDK 11
Plugin log-parser verified successfully with JDK 11 
Collecting metadata for plugin log-parser... Please be patient
Done
Skipping commits changes for plugin log-parser in dry-run mode
Skipping forking plugin log-parser in dry-run mode
Skipping sync plugin log-parser in dry-run mode
Skipping push changes for plugin log-parser in dry-run mode
Skipping pull request changes for plugin log-parser in dry-run mode
*************
Plugin: log-parser
Dry run mode. Changes were made on C:\Users\DELL PC\.cache\jenkins-plugin-modernizer-cli\log-parser\sources but not commited
Modified file: pom.xml
*************
```

### Testing done
Updated the test case to check if ChartUtil is used or not. Also checked that the recipe works on log-parser.

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MigrateStaplerAndJavaxToJakarta.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/MigrateStaplerAndJavaxToJakarta.java
@@ -1,0 +1,109 @@
+package io.jenkins.tools.pluginmodernizer.core.recipes;
+
+import java.util.HashSet;
+import java.util.Set;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.ScanningRecipe;
+import org.openrewrite.TreeVisitor;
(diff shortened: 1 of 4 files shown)


jonesbusy:
Thanks!

jonesbusy (approved):

PR #790, opened 2025-02-10
Feat/add report test coverage module

### LINKED ISSUE #459: Make jacoco work with integration tests

Right now coverage is only computed on unit tests

When we run integration tests with failsafe and 'java -jar..' we should add the argLine to the execution and make sure the execution is merged with previous jacoco execution

### Upstream changes

No code change expected, and maven config

ahmad-kashkoush:
resolves #459 

### Testing done
I've ran 
```sh
mvn clean verify
```
and at the end `plugin-modernizer-report` executes a `report-aggregate` that gather all test coverage from other modules 
image

and when you open `plugin-modernizer-report/target/site/jacoco-aggregate`

you find an HTML file containing all the coverage generated:

image
image
image

### Related links:
- https://www.google.com/search?client=ubuntu-sn&channel=fs&q=report+aggregate
- https://stackoverflow.com/questions/50806424/reporting-and-merging-multi-module-jacoco-reports-with-report-aggregate
- https://www.eclemma.org/jacoco/trunk/doc/report-aggregate-mojo.html
### Submitter checklist


org.jacoco.agent-0.8.12-runtime.jar
jonesbusy, line 1: Do not commit binary files. Maven plugin must be used. The CLI on the issue was just an example

ahmad-kashkoush:
@jonesbusy may give it a review?

org.jacoco.agent-0.8.12-runtime.jar
ahmad-kashkoush: will fix it, anything else?

jonesbusy:
There is many things wrong here

- Coverage is 0 (according on your screenshot). So it's even worse than before. I would expect to see a PR that proof that the computed total coverage is higher if what we already have (60%).
- You commit binary files instead of using the maven jaccoco plugin
- Surefire arg line is never passer to the end process (https://github.com/jenkins-infra/plugin-modernizer-tool/blob/main/plugin-modernizer-cli/pom-it.xml#L39-L41) so it's impossible that we compute coverage when running the final jar.

Like I said on the issue, I removed the 'good-first-issue' tag because it's not and issue for new comers. This require a good understanding on jacoco, maven and how the jar is executed to perform integration test.

Why not trying to implement in in several step, for example

- Ensure the covergage in compute and dump when running test (maybe a different execution dump file
- Ensure the coverage is merged with previous unit execution

Thanks

ahmad-kashkoush:
> There is many things wrong here
> 
>     * Coverage is 0 (according on your screenshot). So it's even worse than before. I would expect to see a PR that proof that the computed total coverage is higher if what we already have (60%).
> 
>     * You commit binary files instead of using the maven jaccoco plugin
> 
>     * Surefire arg line is never passer to the end process (https://github.com/jenkins-infra/plugin-modernizer-tool/blob/main/plugin-modernizer-cli/pom-it.xml#L39-L41) so it's impossible that we compute coverage when running the final jar.
> 
> 
> Like I said on the issue, I removed the 'good-first-issue' tag because it's not and issue for new comers. This require a good understanding on jacoco, maven and how the jar is executed to perform integration test.
> 
> Why not trying to implement in in several step, for example
> 
>     * Ensure the covergage in compute and dump when running test (maybe a different execution dump file
> 
>     * Ensure the coverage is merged with previous unit execution
> 
> 
> 

Ok, thanks for feedback.

Issue #637, opened 2025-01-14
Duplicated tag: 'exclusions'


gounthar:
### Plugin Modernizer version

plugin modernizer 999999-SNAPSHOT (2025-01-14T12:55:12Z)

### What Operating System are you using

Debian Bookworm under WSL2.

### Reproduction steps

`java -jar plugin-modernizer-cli/target/jenkins-plugin-modernizer-999999-SNAPSHOT.jar run --plugins antexec --recipe ReplaceLibrariesWithApiPlugin`

### Expected Results

Build successful.

gounthar:
I was not able to attach the `pom.xml` as a file, so here it is:
```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <groupId>org.jenkins-ci.plugins</groupId>
        <artifactId>plugin</artifactId>
        <version>5.5</version>
        <relativePath />
    </parent>
    <artifactId>antexec</artifactId>
    <version>${changelist}</version>
    <packaging>hpi</packaging>
    <properties>
        <changelist>999999-SNAPSHOT</changelist>
        <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <!-- https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ -->
        <jenkins.baseline>2.479</jenkins.baseline>
        <jenkins.version>${jenkins.baseline}.1</jenkins.version>
    </properties>
    <name>AntExec</name>
    <description>Adds new build step for running directly written Apache Ant code</description>
    <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>

    <developers>
        <developer>
            <id>svasek</id>
            <name>Milos Svasek</name>
            <email>Milos@Svasek.net</email>
            <timezone>+1</timezone>
        </developer>
    </developers>

    <licenses>
        <license>
            <name>The MIT license</name>
            <comments>All source code is under the MIT license.</comments>
            <url>https://opensource.org/licenses/MIT</url>
[...]

PR #791, opened 2025-02-10
Move recipe Replace Libraries With Api Plugin

CodexRaunak:
Issue #637 
Moved the Recipe `ReplaceLibrariesWithApiPlugin` in the plugin-modernizer-tool recipes. Exclusions are not added for libraries that the API plugin handles, as the API plugin is designed to bring those dependencies to the consumer.

### Testing done
Added the `same pom.xml as a test case` where the duplicate tag issue was occurring.

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/recipes/DeclarativeRecipesTest.java
+++ b/plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/recipes/DeclarativeRecipesTest.java
@@ -2587,6 +2587,140 @@ public class DeclarativeRecipesTest implements RewriteTest {
     }
 
     @Test
+    void replaceLibrariesByApiPluginWithBom() {
+        rewriteRun(
+                spec -> spec.recipeFromResource(
+                        "/META-INF/rewrite/recipes.yml",
+                        "io.jenkins.tools.pluginmodernizer.ReplaceLibrariesWithApiPlugin"),
+                // language=xml
+                pomXml(
+                        """
+                <?xml version="1.0" encoding="UTF-8"?>
+                <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+                  <modelVersion>4.0.0</modelVersion>
+                  <parent>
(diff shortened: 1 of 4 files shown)


CodexRaunak:
Need a review on this, 
I added the same pom.xml as a test case where it was adding duplicate tags https://github.com/jenkins-infra/plugin-modernizer-tool/issues/637. The exclusions are not added now at all in `json-api`. But its skipping the exclusions for every library that is a part of API plugin. I am not sure we want that? Also it needs more tests.
 ```
if (isApiProvidedLibrary(groupId, artifactId)) {
                                    continue; // Skip exclusions if the library is already part of the API plugin
  }
```

File: plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/recipes/DeclarativeRecipesTest.java [hunk 1]
```diff
@@ -2586,6 +2586,140 @@ void replaceLibrariesByApiPluginsSimple() throws IOException {
...
+                    <!-- https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ -->
+                    <jenkins.baseline>2.479</jenkins.baseline>
+                    <jenkins.version>${jenkins.baseline}.1</jenkins.version>
+                  </properties>
+                  <dependencyManagement>
+                    <dependencies>
+                      <dependency>
+                        <groupId>io.jenkins.tools.bom</groupId>
+                        <artifactId>bom-${jenkins.baseline}.x</artifactId>
+                        <version>4051.v78dce3ce8b_d6</version>
```
jonesbusy, line 2678: Let's be careful with target bom version since it's moving every week.

Better to use `Settings.getBomVersion()` where possible

jonesbusy:
Hi,

Thanks!

I think we should not exclude anything comming from an API plugin.

Genearally they are used with a BOM, so we should not have dependency convergence issue

I think the situation is much better now than it was originally

plugin-modernizer-core/src/test/java/io/jenkins/tools/pluginmodernizer/core/recipes/DeclarativeRecipesTest.java
CodexRaunak: I think we don't need it here as we are not upgrading the jenkins version and not tracking the weekly release line.

CodexRaunak:
The recipe is not excluding anything from the API plugin.

jonesbusy (approved):

PR #913, opened 2025-04-03
Add Separate Logging for Plugin Modernization Failures

gounthar:
**Please don't merge this pull request, as it is mostly a proof of concept rather than production-ready code.**

Whenever I run a "campaign" (e.g., move to JDK 11, move to JDK 17, migrate to JUnit 5), I encounter numerous failures because lots of plugins just can't migrate.
However, these often go unnoticed as I am processing several hundreds of plugins at once. The failures could stem from the plugin being significantly outdated, limitations in a recipe, bugs in the plugin-modernizer, or issues with a recent version of SpotBugs. Unfortunately, valuable information gets buried within lengthy logs.

The goal of this PR is to create a "log file" that documents all plugin modernization attempts that failed, using the format `plugin:version`. This system will allow us to review details later on (by searching within the local cache for the faulty plugins).
The format also enables us to reuse the list of failed plugins with another recipe (e.g., after failing with JDK 17, some plugins might succeed when moved to JDK 11). Additionally, it allows us to replay the same recipe after making corrections.

In summary, this tool will help me better manage campaigns and hopefully improve the plugin-modernizer overall.

* Add a new appender in `plugin-modernizer-cli/src/main/resources/logback.xml` to log failures separately in `modernization-failures-YYYYMMDD-HHMMSS.log`.
* Update `plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/model/Plugin.java` to log failures in the new log file.
* Add a method in `plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/model/Plugin.java` to log network-related failures in `network-failures-YYYYMMDD-HHMMSS.log`.

### Submitter checklist


jonesbusy:
Which version do we need? Can we use the last published version? It's already fetched 

Yes the current implementation looks fragile

gounthar:
I would need the version we fetched, but I was unable to do so. ;-)

Issue #996, opened 2025-05-07
MigrateToJUnit5 fails for filesystem_scm


gounthar:
### Plugin Modernizer version

1776.vb_3299c07a_d23

### What Operating System are you using

Windows 11

### Reproduction steps

`plugin-modernizer.cmd run --plugins filesystem_scm --recipe MigrateToJUnit5`

### Expected Results

No error, as https://github.com/jenkinsci/filesystem_scm-plugin/pull/24 works and some [earlier PRs](https://github.com/jenkinsci/filesystem_scm-plugin/pull/23) worked today. 🤷 

[invoker.txt](https://github.com/user-attachments/files/20087667/invoker.txt)

gounthar:
It also fails for [codesonar](https://github.com/jenkinsci/codesonar-plugin).

[invoker.txt](https://github.com/user-attachments/files/20089125/invoker.txt)

CodexRaunak:
In the [filesystem_scm-plugin](https://github.com/jenkinsci/filesystem_scm-plugin) it's a compilation err due to this below test

![Image](https://github.com/user-attachments/assets/d234eede-b346-44fa-a172-e9108549edfe)
The above line throws an `Exception` which is not handled, we have 2 options over here either we need to include this line under `assertThrows`, but that may falsify the test to pass or fail if `FileUtils.deleteDirectory(src)` throw an `Exception`, so the best would be we should just `throw the Exception`.
Similar to what this [PR](https://github.com/jenkinsci/filesystem_scm-plugin/pull/24) does.
After:
![Image](https://github.com/user-attachments/assets/047b54cf-8d4c-44dd-9d2a-ddbc29c1c9bc)
 
So we just need to implement a recipe wherever we are using `assertThrows`, for that method we need to throw an `Exception`, as any any line out of `assertThrows` block throwing an `Exception` would cause compile time err.

Working on the recipe ...

jonesbusy:
Fixed

PR #1004, opened 2025-05-15
Fix the hardcoded variable name 'j' of type JenkinsRule in the AnnotateWithJenkins recipe

CodexRaunak:
https://github.com/jenkins-infra/plugin-modernizer-tool/issues/996#issuecomment-2859406207
In the `AnnotateWithJenkins` recipe when we are passing `JenkinsRule j` as a `param` and removing it as a `class variable`, in the recipe we were only looking for the variable named `j`, as it was hardcoded in the recipe. So the recipe would successfully apply for the below case as we have the variable name as `j`. But in case the variable name of `JenkinsRule` is not `j`, it would not pass the `JenkinsRule` as params to the method which would give `cannot be resolved` err and prevents compilation.

### Works Fine When we have JenkinsRule j
Before:
```
public class MyTest {
     @Rule
      public JenkinsRule j = new JenkinsRule();
      @Test
      public void myTestMethod() {
      j.before();
      }
}
```
After:
```
@WithJenkins
class MyTest {
      @Test
      public void myTestMethod(JenkinsRule j) {
      j.before();
      }
}
```

### Error When we have JenkinsRule jr
Before:
```
public class MyTest {
      @Rule
      public JenkinsRule jr = new JenkinsRule();
      @Test
      public void myTestMethod() {
      jr.before();
      }
}
```
After:
```
@WithJenkins
class MyTest {
      @Test
      public void myTestMethod() {   //Not passed JenkinsRule jr as a parameter
      jr.before();   //cannot resolved
      }
}
```

### Testing done
`mvn clean install`

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/AnnotateWithJenkins.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/recipes/AnnotateWithJenkins.java
@@ -6,6 +6,7 @@ import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.ScanningRecipe;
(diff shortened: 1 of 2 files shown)


CodexRaunak:
This is one of the reasons why [codesonar](https://github.com/jenkinsci/codesonar-plugin) was failing as mentioned in https://github.com/jenkins-infra/plugin-modernizer-tool/issues/996#issuecomment-2859406207
There is one more bug, I am working on it.

gounthar:
Nice catch! 🚀

jonesbusy:
Rebased on top of  https://github.com/jenkins-infra/plugin-modernizer-tool/pull/1005

gounthar (approved):

PR #1093, opened 2025-06-28
skip-metadata option to opt out for metadata feature 

CodexRaunak:
- Add option `skip-metadata` with `dry run` and normal `run` commands.
- When skip-metadata is true, it disables collection and pushing it to the repo (both). 
- It is beneficial for testing or development purpose when we don't need to unnecessary add another step of collecting the metadata.
- Updated docs for the cli option
### Testing done
`mvn clean install` Also tested the option with `dry run` and normal `run`, and works as expected.

### Submitter checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,9 @@ source <(plugin-modernizer generate-completion)
 - `--recipe` or `-r`: (required) Name of recipe to apply to the plugins.
 
 
+- `--skip-metadata` (optional) Skip collection and pushing the modernization metadata (i.e metadata after applying the recipes) to the metadata repository. Beneficial for testing or development purpose when we don't need to unnecessary add another step of collecting the metadata.
+
+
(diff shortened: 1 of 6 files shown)


CodexRaunak:
1) Currently when the plugin is `local `we get `modernization metadata` locally but not push it as we don't have scm info.
2) When its `dry run` or a normal `run` we do collect and push both.
3) When we add `skip-metadata` option either with any of the commands `(dry run or run)` , it disables collection and pushing.

I am bit concerned about `dry run`, should we push the metadata in `dry run` state as well or just get it locally and avoid pushing it. 
How about the below:
Dry run - just collect metadata locally, don't push it to the repo
Normal run -  collect and push to the repo
skip-metadata - disable both

jonesbusy:
> How about the below:
> Dry run - just collect metadata locally, don't push it to the repo
> Normal run - collect and push to the repo
> skip-metadata - disable both

Definitely. That's a good idea

Issue #1125, opened 2025-07-15
Failed to get Diff stats in dry-run mode


CodexRaunak:
### Plugin Modernizer version

plugin modernizer 999999-SNAPSHOT (2025-07-15T22:15:58Z)

### What Operating System are you using

Windows 10

### Reproduction steps

`java -jar ./plugin-modernizer-cli/target/jenkins-plugin-modernizer-999999-SNAPSHOT.jar dry-run --plugins buildtriggerbadge -r UpgradeNextMajorParentVersion `

Errored logs:
```
io.jenkins.tools.pluginmodernizer.core.model.PluginProcessingException: Failed to get diff stats
	at io.jenkins.tools.pluginmodernizer.core.model.Plugin.addError(Plugin.java:475)
	at io.jenkins.tools.pluginmodernizer.core.github.GHService.getDiffStats(GHService.java:1773)
	at io.jenkins.tools.pluginmodernizer.core.model.Plugin.getDiffStats(Plugin.java:611)
	at io.jenkins.tools.pluginmodernizer.core.impl.PluginModernizer.collectModernizationMetadata(PluginModernizer.java:476)
	at io.jenkins.tools.pluginmodernizer.core.impl.PluginModernizer.process(PluginModernizer.java:390)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at io.jenkins.tools.pluginmodernizer.core.impl.PluginModernizer.start(PluginModernizer.java:170)
	at io.jenkins.tools.pluginmodernizer.cli.command.DryRunCommand.call(DryRunCommand.java:96)
	at io.jenkins.tools.pluginmodernizer.cli.command.DryRunCommand.call(DryRunCommand.java:19)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2045)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2465)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2457)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2419)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2421)
	at picocli.CommandLine.execute(CommandLine.java:2174)
[...]

PR #1140, opened 2025-07-22
Remove duplicate methods in GH Service and more clean up

CodexRaunak:
- Added an enum `RepoType` that has 2 constants `PLUGIN` and `METADATA`. It defines method calls for that `repoType`.
- Removed all duplicate methods for metadata, making a single method to interact with github (fetch, push, pull, commit etc) based on the `repoType`
- Add exception handling (try-catch) when calculating diff stats due to #1125 
- Clean up for tests 

### Testing done
`mvn clean install` and ran some modernizations on plugin to see everything works as expected.

### Submitter checklist


CodexRaunak:
I think I know why the IT is failing, it's because of the `metadata repo fetch` in dry run mode. Previously we were not doing that, and which is not desirable as well. Because in dry-run mode the idea is not to push or commit any changes, but fetching the repo seems fine. Same we already do for the plugin. 

Therefore I did for the metadata, but seems the IT is failing because of that.

jonesbusy:
Looks fine on main

CodexRaunak:
Changes made:
- Modified the regex of jenkins-version in validation of metadata (fixes failing validations) [#232](https://github.com/Raunak80Madan/metadata-plugin-modernizer/pull/232)
- The IT test was failing due to commiting changes in local repo, now fixed.

jonesbusy (approved):

PR #1222, opened 2025-08-23
Store the opt out repositories and add option to override this behaviour

### LINKED ISSUE #1176: Store the opt-out repositories

Some plugin maintainers choose to reject our automatic PRs, sometimes with an explanation, sometimes without.  
As a result, their plugins remain outdated — still using old Jenkins `parent-pom` versions and core dependencies.

I completely respect their decision, but I don't have the stamina to keep revisiting the same cases. Unfortunately, these plugins will keep receiving new automatic PRs in the next campaign.

Would it be possible to use the existing metadata repository to maintain a list of plugins whose maintainers have opted out of receiving automated PRs?

That way, we can avoid wasting their time and our effort.  
The tool shouldn’t fail in such cases, but instead provide a clear and helpful message explaining the opt-out, along with a link or instructions in case the maintainers wish to opt back in.

Thanks!

### Upstream changes

_No response_

_No response_

CodexRaunak:
Fixes #1176 
- Create enum to store all the plugins that have been opted out, took reference from [here](https://github.com/jenkins-infra/plugin-modernizer-tool/issues/1176#issuecomment-3170701637)
- Add `--override-opt-out-plugins` option to allow making PRs for plugins marked as opt out
### Testing done
made this [PR](https://github.com/jenkinsci/safe-batch-environment-filter-plugin/pull/11) and closed instantly to test the override option

### Submitter checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,9 @@ source <(plugin-modernizer generate-completion)
 - `--skip-metadata` (optional) Skip collection and pushing the modernization metadata (i.e metadata after applying the recipes) to the [metadata repository](https://github.com/jenkins-infra/metadata-plugin-modernizer/). Beneficial for testing or development purpose when we don't need to unnecessary add another step of collecting the metadata.
 
 
+- `--override-opt-out-plugins` (optional) Override the default behavior to allow PRs to plugins marked as opt-out. This forces PR creation regardless of the plugin's opt-out status.
+
+
 - `--clean-forks` (optional) Remove forked repositories before and after the modernization process. Might cause data loss if you have other changes pushed on those forks. Forks with open pull request targeting original repo are not removed to prevent closing unmerged pull requests.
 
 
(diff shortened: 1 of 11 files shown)


plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/model/OptOutPlugins.java
```diff
@@ -0,0 +1,24 @@
+package io.jenkins.tools.pluginmodernizer.core.model;
+
+import java.util.Arrays;
+
+public enum OptOutPlugins {
+    // List of plugins that have opted out from the modernization by plugin-modernizer-tool
+    SAFE_BATCH_ENVIRONMENT_FILTER("safe-batch-environment-filter"),
```
jonesbusy, line 7: I think it would be better to have a file that is not dependent of the release of plugin modernizer

jonesbusy:
> Should we store the opt-out plugins list on the metadata repository, or within the tool like this PR does, if in the metadata, we are required to download it, which maybe an overhead?

It probably the best to have a property file on the metadata repo.

It can be cached using CacheManager to avoid download it each time

Issue #1029, opened 2025-05-26
AddCodeOwner targets a wrong group for openstack-cloud plugin


gounthar:
### Plugin Modernizer version

1837.vf82b_89449a_e7

### What Operating System are you using

Windows 11/PowerShell

### Reproduction steps

```
 plugin-modernizer run --plugins=openstack-cloud --recipe=AddCodeOwner
Run Plugin Modernizer
Skipping verification for plugin openstack-cloud
Fetching plugin code locally openstack-cloud...
Collecting metadata for plugin openstack-cloud... Please be patient
Done
Running recipes io.jenkins.tools.pluginmodernizer.AddCodeOwner for plugin openstack-cloud... Please be patient
Done
Collecting metadata for plugin openstack-cloud... Please be patient
Done
Can only sign commit using SSH key. Skipping sign commit when using GH_TOKEN
Forking plugin openstack-cloud locally from repo openstack-cloud-plugin...
Synced the forked repository for plugin openstack-cloud
Pushed changes to forked repository for plugin openstack-cloud
Pull request created: https://github.com/jenkinsci/openstack-cloud-plugin/pull/417
*************
Plugin: openstack-cloud
Pull request was open on https://github.com/jenkinsci/openstack-cloud-plugin
Modified file: .github/CODEOWNERS
*************

```

### Expected Results

This command lead to https://github.com/jenkinsci/openstack-cloud-plugin/pull/417 .

The content of the generated file is:
`* @jenkinsci/parent-plugin-developers`

PratikMane0112:
@gounthar

`openstack-cloud` plugin has multi-module maven project structure (2 pom.xml files)
- Root level: artifactId: [parent](https://github.com/jenkinsci/openstack-cloud-plugin/blob/d3ba103e31fdf420dda9cccabf7e322251bc6c65/pom.xml#L12)
- Plugin subdirectory: artifactId: [openstack-cloud](https://github.com/jenkinsci/openstack-cloud-plugin/blob/d3ba103e31fdf420dda9cccabf7e322251bc6c65/plugin/pom.xml#L11)

The recipe works as designed by generating team names in the format `@jenkinsci/{artifactId}-plugin-developers` The difference in results appears to be related to which `pom.xml` file is being used:

- `--plugins=openstack-cloud` option references root pom.xml and results : `@jenkinsci/parent-plugin-developers`
[...]

PR #1459, opened 2026-01-09
Better support for multi-module plugins

PratikMane0112:
Fix issue #1029 

Used logic -
- for single module plugin :
    - Root POM has `packaging=hpi`,  will refer root directory 
    - No change in behavior
- for multi module plugin :
    - Root POM has packaging=pom then search subdirs
    - Find subdir with `packaging=hpi`
    - Use that subdir's POM 

### Testing done

With this logic I tried with below plugins & was able to add correct teams

- --plugins option (local .cache)
```java
plugin-modernizer --plugins openstack-cloud --recipes AddCodeOwner
```

- --plugin-path (using test-plugins dir)
```java
plugin-modernizer --plugin-path test-plugins/openstack-cloud --recipes AddCodeOwner
```

dry-run recipe over other plugins: [badge](https://plugins.jenkins.io/badge/), [git](https://plugins.jenkins.io/git/) plugins (single module) & [workflow-cps](https://github.com/jenkinsci/workflow-cps-plugin) (multi-module)

The content  added to `.github/CODEOWNERS` w.r.t plugins -

- openstack-cloud-plugin : `* @jenkinsci/openstack-cloud-plugin-developers`
- badge-plugin : `* @jenkinsci/badge-plugin-developers`
- git-plugin : `* @jenkinsci/git-plugin-developers`
- workflow-cps-plugin : `* @jenkinsci/workflow-cps-plugin-developers`

### Submitter checklist

### DIFF
--- a/homebrew/formula.rb.tpl
+++ b/homebrew/formula.rb.tpl
@@ -10,7 +10,7 @@ class PluginModernizer < Formula
 
     def install
       libexec.install "jenkins-plugin-modernizer-{{projectVersion}}.jar"
-      bin.write_jar_script libexec/"jenkins-plugin-modernizer-{{projectVersion}}.jar", "plugin-modernizer"
+      bin.write_jar_script libexec/"jenkins-plugin-modernizer-{{projectVersion}}.jar", "plugin-modernizer", "--add-opens=java.base/java.lang=ALL-UNNAMED --sun-misc-unsafe-memory-access=allow"
     end
 
     test do
--- a/jreleaser.yml
+++ b/jreleaser.yml
@@ -10,7 +10,7 @@ project:
     homepage: https://github.com/jenkins-infra/plugin-modernizer-tool
   java:
     groupId: io.jenkins.plugin-modernizer
-    version: 17
+    version: 25
   inceptionYear: 2024
 
 release:
--- a/plugin-modernizer-core/pom.xml
+++ b/plugin-modernizer-core/pom.xml
@@ -311,6 +311,17 @@
                   <artifactId>hamcrest</artifactId>
                   <version>3.0</version>
                 </dependency>
+                <!-- JSR305 annotations -->
+                <dependency>
+                  <groupId>javax.annotation</groupId>
+                  <artifactId>javax.annotation-api</artifactId>
+                  <version>1.3.2</version>
+                </dependency>
+                <dependency>
+                  <groupId>com.google.code.findbugs</groupId>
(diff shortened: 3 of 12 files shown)


PratikMane0112:
@jonesbusy @gounthar, I had used a simple logic for detection & adding of correct team as `CODEOWNER` for plugin,  
if any other logic which is better, would also be applied & really appreciated :)

plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/converter/PluginPathConverter.java
```diff
@@ -2,36 +2,97 @@
...
-        StaticPomParser staticPomParser = new StaticPomParser(pom.toString());
-        String packaging = staticPomParser.getPackaging();
-        if (!"hpi".equals(packaging)) {
-            throw new IllegalArgumentException("Path does not contain a Jenkins plugin: " + path);
+
+        StaticPomParser rootPomParser = new StaticPomParser(pom.toString());
+        String packaging = rootPomParser.getPackaging();
+
+        // Check if this is a single-module Jenkins plugin
+        if ("hpi".equals(packaging) || "jenkins-plugin".equals(packaging)) {
```
jonesbusy: Any example of packaging type "jenkins-plugin"?

AI hallucination?

plugin-modernizer-cli/src/main/java/io/jenkins/tools/pluginmodernizer/cli/converter/PluginPathConverter.java
```diff
@@ -2,36 +2,97 @@
...
 
+    private static final Logger LOG = LoggerFactory.getLogger(PluginPathConverter.class);
+
     @Override
     public Plugin convert(String value) throws Exception {
         Path path = Path.of(value);
         if (!Files.isDirectory(path)) {
             throw new IllegalArgumentException("Path is not a directory: " + path);
         }
-        // We assume the plugin is at the pom, will not work for multi module projects
```
jonesbusy: If I understand correctly we return the subpath that contains only the plugin pom?

Are we sure we are able to compile/modernize such plugin without doing it from parent pom?

Perhaps it's worth to have an integration test with such layout

jonesbusy:
Some comments to address.

Also an integration test would be useful (CommandLineIT)

PR #1655, opened 2026-03-21
Use except Exception instead of bare except in validate_metadata.py

Fikri-20:
Bare `except:` clauses catch everything including `KeyboardInterrupt` and `SystemExit`, which should never be silently caught.

Changed both occurrences in `validate_metadata.py` to `except Exception:` to only catch actual errors.


jonesbusy:
@CodexRaunak Would you help me to review? Thanks!

PR #1657, opened 2026-03-23
Fix/cross platform ci matrix

jay7-tech:
Added an OS matrix [ubuntu-latest, windows-latest, macos-latest] to the modernizer action workflow.

Currently, the action is only tested on Ubuntu. However, since Jenkins plugins are heavily developed on Windows/macOS, the underlying 

action.yml
 (which uses raw bash pathing like $(which mvn) | sed...) will likely break on those runners. This matrix surfaces those hidden cross-platform bugs automatically.

Testing done
Verified workflow syntax locally. The automated PR checks will now natively spin up Windows/macOS runners to prove if the action works cross-platform.

Submitter checklist
 [x] Make sure you are opening from a topic/feature/bugfix branch (right side) and not your main branch!
 [x] Ensure that the pull request title represents the desired changelog entry
 [x] Please describe what you did
 [x] Link to relevant issues in GitHub or Jira (N/A)
 [x] Link to relevant pull requests, esp. upstream and downstream changes (N/A)
 [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/.github/actions/modernizer-action/action.yml
+++ b/.github/actions/modernizer-action/action.yml
@@ -37,7 +37,7 @@ runs:
       shell: bash
       working-directory: ${{ inputs.working-directory }}
       run: |
-        if [ ${{ inputs.version }} == "latest" ]; then
+        if [ "${{ inputs.version }}" = "latest" ]; then
           API_URL="https://api.github.com/repos/jenkins-infra/plugin-modernizer-tool/releases/latest"
         else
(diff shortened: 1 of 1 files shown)


jonesbusy:
Why this PR contains code change? I'm not sure it's worth to test on Windows or MacOSX for the action.

The action is only meant to be run on GitHub action runner (Linux, so reason to choose different)

PR #1696, opened 2026-04-12
fix: three silent-failure bugs in verification, cache path, and JDK selection

Fikri-20:
Three silent bugs I found while working on the tool. Each one causes the tool to proceed as if nothing
  went wrong when it should have stopped or warned.                                                     
                                                   
  ### 1. Verification failures were silently swallowed (PluginModernizer.verifyPlugin)
                                                                                                            
  `verifyPlugin()` called `plugin.withoutErrors()` unconditionally at the end, regardless of whether the build  actually passed. The caller in `start()` would always see a clean plugin and continue to fork, commit, and  open a PR — even when the verification build failed. Removed both `withoutErrors()` calls and upgraded the log from INFO to WARN.                                                                                  

In dry-run mode, the error is still cleared after logging so the diff preview is still shown to the user (that's intentional — you're previewing, not committing).
                                                                                                            
###  2. `getDiffStats` always read from the default cache path (`GHService`)                                     

  `getDiffStats()` was resolving the .git directory against Settings.DEFAULT_CACHE_PATH instead of            
  `config.getCachePath()`. So running with `--cache-path /tmp/my-cache` would clone the plugin correctly there
  but then fail to read it back when building the PR description. One-line fix.                             
                                                                                                          
###  3. `JDK.min(Set, String)` ignored the `jdks` argument (JDK)

  Operator precedence bug in the guard:

  // was parsed as:
  if `(jdks == null || (jdks.isEmpty() && jenkinsVersion == null))`
                                                                                                            
So a non-null, non-empty jdks fell straight through to `JDK.get(jenkinsVersion)` with no filtering — jdks   was dead code. A plugin that declares [17, 21] in its Jenkinsfile would get compiled with Java 11 against an older Jenkins version. Rewrote the method to intersect the declared JDKs with the version-supported set and added tests for the untested paths.                                                                

  ---
  Testing: all three reproduced locally before the fix, gone after. mvn test -pl plugin-modernizer-core
  passes.

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/github/GHService.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/github/GHService.java
@@ -1237,7 +1237,7 @@ public class GHService {
      * @return DiffStats (no. of additions, deletions and changed files)
      */
     public DiffStats getDiffStats(Plugin plugin, boolean dryRun) {
-        Path gitDirPath = Settings.DEFAULT_CACHE_PATH
+        Path gitDirPath = config.getCachePath()
                 .resolve(plugin.getName())
                 .resolve("sources")
                 .resolve(".git")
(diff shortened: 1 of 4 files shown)


jonesbusy:
If this fixes 3 bug please open 3 PR.

While I agree with 1 and 2 I would like to study more the 3rd

Opening a PR with different contexts reduce the change it gets merged or reviewer

Fikri-20:
yeah I see, breaking them into 3 separate PRs now ..

Fikri-20:
done, #1711, #1712, #1713

PR #1708, opened 2026-04-15
feat: record failure reasons in modernization metadata

ap44444:
When a plugin migration fails, the tool writes "migrationStatus": "fail" to the modernization metadata JSON but provides no explanation. So 46% of all records in the metadata repository are failures with no descriptive information and maintainers cannot tell whether the failure was due to a precondition (e.g. parent POM 1.x, HTTP repositories) or runtime build error.

This PR adds a failureReasons field to ModernizationMetadata that takes the error messages at the time of failure. Precondition errors (e.g. "Found parent version starting with 1. in pom file preventing modernization") are recorded first, and then any runtime exception messages. This field is not there in JSON entirely when the migration succeeds, so existing successful records will not be affected.

### Testing done

Two unit tests added to PluginModernizerTest:

- testCollectModernizationMetadata_WithPreconditionErrors_ShouldRecordFailureReasons: this verifies that a precondition  failure (PARENT_POM_1X) is captured in failureReasons and migrationStatus is "fail"
- testCollectModernizationMetadata_WithRuntimeErrors_ShouldRecordFailureReasons: this verifies that a runtime build error message is captured in failureReasons

### Submitter checklist

### DIFF
--- a/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/ModernizationMetadata.java
+++ b/plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/ModernizationMetadata.java
@@ -82,6 +82,12 @@ public class ModernizationMetadata extends CacheEntry<ModernizationMetadata> {
     private String migrationStatus;
 
     /**
+     * Reasons why the migration failed, if any.
+     * Null when migration succeeded (omitted from JSON).
+     */
+    private List<String> failureReasons;
+
+    /**
      * Number of deprecated APIs removed by the migration
      */
     private Integer removedDeprecatedApis;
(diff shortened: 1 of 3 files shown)


plugin-modernizer-core/src/main/java/io/jenkins/tools/pluginmodernizer/core/extractor/ModernizationMetadata.java
```diff
@@ -81,6 +81,12 @@ public class ModernizationMetadata extends CacheEntry<ModernizationMetadata> {
      */
     private String migrationStatus;
 
+    /**
+     * Reasons why the migration failed, if any.
+     * Null when migration succeeded (omitted from JSON).
+     */
+    private List<String> failureReasons;
```
jonesbusy, line 88: I'm not sure a list is enough. Each execution will clear result.

I'm wonder if we need a Map<String, String> instea (recipe/failure)

I need to think more about it

ap44444: Thank you, I think that would be helpful too, but I chose a list because the current codebase does not have error tracking for each recipe. The error models 'PreconditionError and PluginProcessingException' returns lists of errors for each plugin, but do not specify to which sub-recipe. A map may require changes to how the errors are tracked.

